PR #3, opened 2021-08-05
difftest: fix snapshot APIs

poemonsense:


AugustusWillisWang (approved):

AugustusWillisWang:
Current snapshot does not contain store queue in NEMU, which will be added later.

PR #51, opened 2021-11-10
Implement difftest modules in Chisel ExtModule

poemonsense:


wakafa1 (dismissed):

src/main/scala/Difftest.scala
```diff
@@ -167,64 +168,135 @@ class DiffRunaheadMemdepPredIO extends DifftestBundle with DifftestWithIndex {
...
+
+  def getDPICArgString(argName: String, data: Data): String = {
+    val dpicArgTypes = Seq(
+      ((width: Int) => width == 1)                 -> "bit",
+      ((width: Int) => (2 to 8).contains(width))   -> "byte",
+      ((width: Int) => (9 to 32).contains(width))  -> "int",
+      ((width: Int) => (33 to 64).contains(width)) -> "longint"
+    )
+    val typeString = dpicArgTypes.filter(_._1(data.getWidth)).map(_._2)
+    require(typeString.length == 1, s"unsupported io type of width ${data.getWidth}!!\n")
```
ljwljwljwljw, line 182: A `match case` may make code more readable.

ljwljwljwljw (changes requested):

src/main/scala/Difftest.scala
ljwljwljwljw: eg:
```
witdh match {
    case 1 => "bit"
    case byte if byte >=2 && byte <=8 => "byte"
    ...
}
```

src/main/scala/Difftest.scala
```diff
@@ -167,68 +168,144 @@ class DiffRunaheadMemdepPredIO extends DifftestBundle with DifftestWithIndex {
...
+  def getDPICArgString(argName: String, data: Data): String = {
+    val directionString = getDirectionString(data)
+    val typeString = data.getWidth match {
+      case 1                                  => "bit"
+      case width if width > 1  && width <= 8  => "byte"
+      case width if width > 8  && width <= 32 => "int"
+      case width if width > 32 && width <= 64 => "longint"
+      case _ => s"unsupported io type of width ${data.getWidth}!!\n"
+    }
+    val argString = Seq(directionString, f"${typeString.head}%7s", argName)
```
ljwljwljwljw, line 187: Is `head` a typo in `${typeString.head}`?

ljwljwljwljw (changes requested):

ljwljwljwljw (approved):

PR #56, opened 2021-12-14
Support difftest in debug mode

lqre:


src/main/scala/Difftest.scala
```diff
@@ -61,6 +61,7 @@ class DiffBasicInstrCommitIO extends DifftestBundle with DifftestWithIndex {
 class DiffInstrCommitIO extends DiffBasicInstrCommitIO {
   val pc       = Input(UInt(64.W))
   val instr    = Input(UInt(32.W))
+  val scFailed = Input(Bool())
```
AugustusWillisWang, line 64: `scFailed` has been replaced with `DiffLrScEventIO`, see #55

poemonsense (approved):

PR #69, opened 2022-04-06
api: support generate verilator emulator dl

AugustusWillisWang:


src/test/csrc/difftest/refproxy.cpp
```diff
@@ -88,17 +88,17 @@ NemuProxy::NemuProxy(int coreid) {
...
     nemu_difftest_set_mhartid(coreid);
-  }
 
-  auto nemu_misc_put_gmaddr = (void (*)(void*))dlsym(handle, "difftest_put_gmaddr");
-  if (NUM_CORES > 1) {
-    check_and_assert(nemu_misc_put_gmaddr);
-    assert(goldenMem);
-    nemu_misc_put_gmaddr(goldenMem);
+    auto nemu_misc_put_gmaddr = (void (*)(void*))dlsym(handle, "difftest_put_gmaddr");
+    if (NUM_CORES > 1) {
```
poemonsense, line 97: no need for this `if`?

poemonsense:
I'm wondering why do we need this?

verilator.mk
```diff
@@ -84,6 +84,12 @@ ifeq ($(RELEASE),1)
 EMU_CXXFLAGS += -DBASIC_DIFFTEST_ONLY
 endif
 
+ifeq ($(GEN_DYNAMIC_LIB), 1)
+EMU_CXXFLAGS += -shared -fPIC
+else
+EMU_CXXFLAGS += -fPIE
```
poemonsense, line 90: why not put `-fPIE` at line 25?

src/test/csrc/difftest/refproxy.cpp
AugustusWillisWang: No need, will be fixed.

AugustusWillisWang:
> I'm wondering why do we need this?

Another team needs to use an RTL-level RISC-V core simulator in a cpp project. By adding dl interface, they can easily run core with difftest in their project.

As difftest is a little bit messy for now, this feature will not be merged. It will stay in an independent branch.

PR #77, opened 2022-05-02
fix: do not do_store_check before FIRST_INST_ADDRESS

AugustusWillisWang:
REF model is not setuped until the 1st inst commit. do_store_check before
that will always fail

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -100,19 +100,21 @@ int Difftest::step() {
   if (check_timeout()) {
     return 1;
   }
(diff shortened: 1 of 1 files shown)


poemonsense:
A correct design should not have any store before an instruction commits. But a wrong design may cause wrong values to be stored into memory before any instruction commits.

AugustusWillisWang:
What if we need to run some code (including store) before difftest begins? For example, a loader in flash?

poemonsense:
Difftest is enabled during the flash. Moving store_check after has_commit will cause stores to be ignored.

PR #97, opened 2023-01-08
difftest: Add PTW refill check

good-circle:
Add PageCache_refill_check which is similar to icache_refill_check and dcache_refill_check. I have checked that both icache_refill_check and dcache_refill_check can still work after this commit.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   test-difftest:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
       - uses: actions/checkout@v2
--- a/src/main/scala/Difftest.scala
+++ b/src/main/scala/Difftest.scala
@@ -170,7 +170,7 @@ class DiffRefillEventIO extends DifftestBundle {
   val valid = Input(Bool())
   val addr  = Input(UInt(64.W))
   val data  = Input(Vec(8, UInt(64.W)))
-  val cacheid = Input(Bool())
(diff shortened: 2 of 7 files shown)


wakafa1 (dismissed):

wakafa1:
Given PTW refill test, should we no longer need to limit the occurrence of force_raise_pf?

Lemover:
> Given PTW refill test, should we no longer need to limit the occurrence of force_raise_pf?

force_raise_pf is still needed.
force_raise_pf solve the problem: software has changes the mmu page into the state that will not raise pf, but lazily sfence mmu, for higher performance. When XS has the stale mmu,  XS and NEMU will execute at different way, raise pf and not raise pf. force_raise_pf then works.

The ptw_refill_test solves another different problem that Cache System(include store commit buffer) has an error. In the force_raise_pf case, mmu should have the stale entry, which means mmu will hit and no ptw req is needed.

wakafa1:
> > Given PTW refill test, should we no longer need to limit the occurrence of force_raise_pf?
> 
> force_raise_pf is still needed. force_raise_pf solve the problem: software has changed the mmu page into the state that will not raise pf, but lazily sfence mmu, for higher performance. When XS has the stale mmu, XS and NEMU will execute at different way, raise pf and not raise pf. force_raise_pf then works by forcing NEMU to have a pf.
> 
> The ptw_refill_test solves another different problem that Cache System(include store commit buffer) has an error. In the force_raise_pf case, mmu has the stale entry, which means mmu will hit and no ptw req is needed.

Agree. The mechanism should always behave, but I think we no longer need to report an error if occurrence of forced PF exceeds certain number, say 5.

wakafa1 (approved):

PR #113, opened 2023-03-15
diffRefillEvent: make it more flexible

Guo-HY:
Use an array to store the refill structure

add idtfr signal in INTERFACE_REFILL_EVENT dpic func for flexible usage


chenguokai (dismissed):

chenguokai (approved):

Guo-HY:
changes：

Use an array to store the refill structure

add idtfr signal in INTERFACE_REFILL_EVENT dpic func for flexible usage

do_refill_check : continue run some cycle before aborted to dump wave

PR #126, opened 2023-08-10
verilator: add support for Verilator 4.228

ptomsich:
We can check VERILATOR_VERSION_INTEGER from include/verilated_config.h for the numeric Verilator version (e.g., 4.228 becomes 4228000), so there is no need for our own symbolic macros (e.g., VERILATOR_4_210).

This commit:
- replaces the usage of our newly defined macros and directly tests the VERILATOR_VERSION_INTEGER macro
- adds support for the VlThreadPool(...) signature used in 4.228 and newer of Verilator


poemonsense:
I agree with you on the usage of `VERILATOR_VERSION_INTEGER` macro and the `VlThreadPool`.

However, there is one concern on the support for the v4.218 version we are localling using and recommending the users to use as well. As seen in https://github.com/verilator/verilator/issues/3343, the new macro was introduced in v4.220. So the check `elif VERILATOR_VERSION_INTEGER >= 4210000` will not be effective if the version is older than v4.220.

I'm assuming we will upgrade verilator directly to v5.016 soon and, at that time, we can completely remove `VERILATOR_4_210`. So this time I would perfer to leave the `VERILATOR_4_210` macro unchanged, while other changes look good to me.

Would you mind change these two lines back to the original version? `#elif VERILATOR_VERSION_INTEGER >= 4210000` and `EMU_CXXFLAGS += -DVERILATOR_4_210`. We are happy to approve this PR if they are changed back.

Also, but a minor one, what's this line in Makefile for?
```
VERILATOR_VERSION = $(shell `$(VERILATOR_VER_CMD)`)
```

poemonsense (approved):

PR #157, opened 2023-09-18
Support overwriting the gcpt restorer manually

xyyy1420:


src/test/csrc/verilator/emu.cpp
shinezyy, line 1: overwrite_nbytes` should not use 'o' as shorthand

shinezyy (changes requested):
overwrite_nbytes should not use 'o' as shorthand, because o is usually used for output or some common options.

please remove 'o' from L161, L172, and L263

shinezyy (dismissed):
seems Ok

src/test/csrc/verilator/emu.cpp
```diff
@@ -83,6 +84,7 @@ static inline void print_help(const char *file) {
   printf("  -R, --ipc-interval=NUM     the interval insts of drawing IPC curve\n");
 #endif
   printf("  -X, --fork-interval=NUM    LightSSS snapshot interval (in seconds)\n");
+  printf("      --overwrite-nbytes=n   set valid bytes,but less than 0xf00,default value is 0x700\n");
```
poemonsense, line 87: The default value here is different from the value in `uint64_t overwrite_nbytes = 0xe00;`? 

Better to use `default: 0xe00` to reduce the length of the message and add a whitespace after `,`

poemonsense (dismissed):

src/test/csrc/verilator/emu.cpp
```diff
@@ -368,6 +374,12 @@ Emulator::Emulator(int argc, const char *argv[]):
     }
   }
 
+  if (args.gcpt_restore) {
+    InputReader *reader=new FileReader(args.gcpt_restore);
```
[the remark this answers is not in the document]
poemonsense: and a whitespace after `,`

PR #167, opened 2023-09-21
Avoid include Verilator files in src/difftest

poemonsense:


poemonsense:
will rebase over #169

PR #174, opened 2023-09-22
refproxy: fix the first argument of debug_mem_sync

eistar:


poemonsense:
I agree with the changes as they follow the interfaces of Spike. However, it seems like the declaration of `RefProxy::guided_exec` should be changed as well to avoid compiler errors. The CI fails because of the errors.

BTW, I think we need to update the legacy NEMU interface for `difftest_guided_exec` with a return value. Otherwise if in the future someone uses the return value to decide something, NEMU may return errors.

eistar:
Thanks for the explanation. I reverted the change to difftest_guided_exec to prevent affecting Difftest with NEMU at this monent.

poemonsense:
Thanks. I agree we should upgrade NEMU in the future, probably in a separated PR

poemonsense (approved):

PR #175, opened 2023-09-24
emu: delete lightsss to avoid resource leaks

poemonsense:
Shared memory resources should be released manually. Otherwise the system would be out of shared memory resources.


poemonsense:
4850feb825cf374309b5c6a009f109a5d435e08b fixes the resource leak but 4c1b283b420f09428caed537f16a06af53526ba9 introduces it again. We need to fix it, otherwise the simulation fails immediately if the system is full of unreleased share memory.

PR #188, opened 2023-10-16
Difftest: enable batch mode

klin02:
Temp commit: successfully load trace and run n steps in single cycle.
Cannot pass CI now! And still some dirty code left:)
I will try to store n trace in a single cycle later~


klin02:
Enable vcs with batch later，and ci will be added. Do not merge now

klin02:
It looks like verilator 5.016 has a potential bug that doesn't handle a large number of flying lines, but verilator 5.012 works.

We may use Verilator 5.012 for this PR, and revert to 5.016 after pack up flying lines. See #206

poemonsense (approved):

PR #202, opened 2023-10-23
Delay difftest_step by one cycle

klin02:
To ensure diffteststate obtained by software is after DPIC. 
Use dut_ways and select to avoid missing state written by DPIC


klin02:
When `make emu`, we ensure that Difftest read DiffState after DPIC calls in the same beat.

However when `make simv` with VCS，the order of DPIC calls and difftest steps in the same beat is random, so we need to delay step_func by 1 clk. 

The operation of delaying is only needed by VCS.

klin02:
Still need to share wb_int betweens different duts. I will fix it later~

klin02:
The lastest commits based on #204, because it will more clear to handle wb_int under the new code organization.

This PR should be merged after #204

klin02:
Function of this PR has been implemented in #188. Therefore, this PR can be closed.

PR #204, opened 2023-10-27
Gateway: handles logic independent to DPIC

klin02:
* Gateway: handles logic independent to specific transfer way, such as DPIC...
* DPIC: focus on DPIC interfaces, only deal with code generation.

By separating the logic, we can pave the way for other transfer way. And modifications not related to DPIC will all be made on the Gateway.


klin02:
Function of this PR has been implemented in #188. Therefore, this PR can be closed.

PR #210, opened 2023-11-20
Difftest: enable Squash Replay

klin02:
Replay the unsquash trace corresponding to the failed squashed trace, for more accurately locating the error


klin02:
Feature List:
* Move the squash to the gateway and remove useless fly-lines. 
* Supports Squash snapshot and replay, use Memcpy to snapshot Ref Mem, need to be optimized later.
* Add Control module for squash_idx(Signals like Step will be added later), derived from DPICBase
* Gen DiffStateBuffer virtual declaration along with the implementation, so we can add other singals easier.

src/main/scala/Difftest.scala
```diff
@@ -99,7 +98,7 @@ sealed trait DifftestBundle extends Bundle with DifftestWithCoreid { this: Difft
   }
 
   // returns Bool indicating whether `this` bundle can be squashed with `base`
-  def supportsSquash(base: DifftestBundle): Bool = supportsSquashBase
+  def supportsSquash(base: DifftestBundle, maxNumFused: Int): Bool = supportsSquashBase
```
poemonsense, line 101: what's this for?

src/main/scala/Difftest.scala
```diff
@@ -120,8 +119,7 @@ class DiffInstrCommit(nPhyRegs: Int = 32) extends InstrCommit(nPhyRegs)
 {
   override val desiredCppName: String = "commit"
 
-  private val maxNumFused = 255
-  override def supportsSquash(base: DifftestBundle): Bool = {
+  override def supportsSquash(base: DifftestBundle, maxNumFused: Int): Bool = {
```
poemonsense, line 122: what's this for?

poemonsense (commented):
It seems too many commits are mixed into this large PR. It is difficult for me to review the code.

I would suggest spliting the modifications so that I can correct review the changes. But if you insist on making them a large PR, it's ok to merge it

src/main/scala/Difftest.scala
klin02: It's to use config in Gateway.scala to decide squashSize. 
We want all related params to be specified in the same case class, so that the params can be combined and changed easily.
And I pass config.squashSize -1 as maxNumFused in squash.scala

klin02:
> It seems too many commits are mixed into this large PR. It is difficult for me to review the code.
> 
> I would suggest spliting the modifications so that I can correct review the changes. But if you insist on making them a large PR, it's ok to merge it

The feature list above summarizes the contains of this PR. 
* To place related params together, we move instantiation of Squash into Gateway, and then remove useless fly-lines because GatewayEndpoint has finished it.
* To support Squash replay, we need corresponding hw and sw implementation. And to make it configurable, we must add related singals in `DiffStateBuffer` and a control module to pass squash_idx flexibly, it's hard to clearly split it.
* For configurability, we move virtual Class Declaration inside DPIC and pass it to Difftest. And the control module (DPICControl, now used to pass squash_idx) deprived from DPICBase like other DPIC module. It will help us add other control signals easier. Now this module will be instantiated in the same DummyDPICWrapper as DiffArchEvent (use isUniqueIdentifier as cond, to make sure it will instaniatiated once in each core). Most of code changes come from this and hard to split them.

src/main/scala/Difftest.scala
poemonsense: The definition of supportsSquash has nothing to do with `maxNumFused`. `maxNumFused` is only for InstrCommit bundle.

poemonsense:
> > It seems too many commits are mixed into this large PR. It is difficult for me to review the code.
> > I would suggest spliting the modifications so that I can correct review the changes. But if you insist on making them a large PR, it's ok to merge it
> 
> The feature list above summarizes the contains of this PR.
> 
> * To place related params together, we move instantiation of Squash into Gateway, and then remove useless fly-lines because GatewayEndpoint has finished it.
> * To support Squash replay, we need corresponding hw and sw implementation. And to make it configurable, we must add related singals in `DiffStateBuffer` and a control module to pass squash_idx flexibly, it's hard to clearly split it.
> * For configurability, we move virtual Class Declaration inside DPIC and pass it to Difftest. And the control module (DPICControl, now used to pass squash_idx) deprived from DPICBase like other DPIC module. It will help us add other control signals easier. Now this module will be instantiated in the same DummyDPICWrapper as DiffArchEvent (use isUniqueIdentifier as cond, to make sure it will instaniatiated once in each core). Most of code changes come from this and hard to split them.

I'll approve this PR. You can merge this PR as you like. But someone will have to spend more time understanding and optimizing the code structure in the future. I doubt the current code structure is clear enough as a Scala project.

poemonsense (approved):

poemonsense:
Please squash and merge the PR. Do not simply merge it with a merge commit.

PR #218, opened 2023-11-24
Fix: replace pointer to instance array with pointer array

klin02:
Parent class pointer increment will still use parent class size when pointing to derived class instance


klin02:
When run simv with multi-cores, there is still some segmentation fault as follows. Need to be fixed later
1124bug

poemonsense (approved):

PR #219, opened 2023-11-27
Add support for common DifftestTopIO

poemonsense:
* Add LogPerfControl
* Force all DiffTest IOs to have prefix of `difftest_`


poemonsense:
It seems Perf and Log IOs are coupled with the simulation framework in DiffTest. We should move them from utility to DiffTest to allow anyone who do not use utility but want to use the log/perf interfaces. Probably we can migrate the XiangShan log/perf interface code to this new Top IO provided by Difftest. @Tang-Haojin 

I'm now using difftest like this:

For customized perf/log IO:
```scala
object PerfAccumulate {
  def apply(perfName: String, perfCnt: UInt) = {
    val control = LogPerfControl()  // call DiffTest.LogPerfControl to get the control bits from difftest

    val counter = RegInit(0.U(64.W)).suggestName(perfName + "Counter")
    val next_counter = WireInit(0.U(64.W))
    next_counter := counter + perfCnt
    counter := Mux(control.clean, 0.U, next_counter)

    when (control.dump && control.logEnable) {
      printf(p"[cycle=${control.timer}] $perfName: $next_counter\n")
    }
  }
}
```

For the top (SimTop):
```scala
    val difftest = DifftestModule.finish("Piper") // It will instantiate the top IOs to interact with DiffTest
    difftest.uart <> mmio.io.uart
```

Tang-Haojin (approved):
Sure. Let me handle it.

klin02 (approved):

PR #221, opened 2023-12-06
Use dut_pos to replace select or batch_idx

klin02:


src/main/scala/DPIC.scala
```diff
@@ -209,16 +189,19 @@ object DPIC {
...
-      else 1
-    }
-    val class_def =
+
+    val interfaceCpp = ListBuffer.empty[String]
+    interfaceCpp += "#ifndef __DIFFTEST_DPIC_H__"
+    interfaceCpp += "#define __DIFFTEST_DPIC_H__"
+    interfaceCpp += ""
+    interfaceCpp += "#include <cstdint>"
+    interfaceCpp += "#include \"diffstate.h\""
```
poemonsense, line 198: It seems these lines do not need to change.

src/main/scala/DPIC.scala
```diff
@@ -251,8 +226,17 @@ object DPIC {
     val outputHeaderFile = outputDir + "/difftest-dpic.h"
     Files.write(Paths.get(outputHeaderFile), interfaceCpp.mkString("\n").getBytes(StandardCharsets.UTF_8))
 
-    val diff_func =
+    interfaceCpp.clear()
+    interfaceCpp += "#ifndef CONFIG_NO_DIFFTEST"
+    interfaceCpp += ""
+    interfaceCpp += "#include \"difftest.h\""
```
poemonsense, line 232: These lines are not necessary to change as well

src/main/scala/Difftest.scala
```diff
@@ -401,8 +401,6 @@ object DifftestModule {
          |extern void diffstate_buffer_init();
          |extern void diffstate_buffer_free();
          |""".stripMargin
-
-    difftestCpp += class_def
```
poemonsense, line 405: We should avoid changes like this.

If you decide to optimize the coding style, make it a separate PR. This PR is for the dut_pos functional change only. Please do not let more changes in.

Coding style changes usually come with a large number of lines changed. They should not be counted in this functional change PR.

poemonsense (commented):
Also write some commit/PR messages on GitHub to explain why and how you propose this PR please. Git messages would help you understand the code in the future. After several months, everyone will be unable to remember why they made the changes before.

poemonsense (approved):

poemonsense:
Feel free to squash and merge it

PR #231, opened 2023-12-27
misc: expect rtl to be created at BUILD_DIR/rtl

poemonsense:


poemonsense:
CIRCT has some wierd settings, such as the codegen will put every source file into one large file, including the non-RTL source files such as blackbox.f. Therefore, we have to separate these files manually.

Now I'm facing a problem with `--export-chisel-interface`. This option creates a scala file in the beginning of the generated Verilog  and we can hardly separate it from the large verilog. CIRCT really does bad on the codegen.

One better solution (also preferred when creating release Verilog) is to use `--split-verilog`. However, this would make all files separated and create a lot of files. Therefore, I propose to move all files to the rtl directory. This would benefit everyone with better clarity for chisel users.

klin02 (approved):

PR #235, opened 2023-12-29
wiring: disabled by default

poemonsense:
We provide several functions to create IOs or wires via DifftestWiring.

* remove extra IO support in Difftest.finish
* separate createPendingWires and createExtraIOs


poemonsense:
We allow both top-level IOs or blackboxes for the extra IOs. This improves the flexibility.

PR #242, opened 2024-01-12
SimJTAG: use logic for random

klin02:
Palladium doesn't support set $random to wire. Use dummy logic val to set it.


poemonsense (changes requested):
Commit title should explain itself, i.e., what the change is. This PR should be about using logic for random.

Second, I think there is no need to switch between the two code snippets here. Logic random also applies to other simulators. So we can use the new logic one only.

poemonsense (approved):

PR #245, opened 2024-01-12
Makefile,Scripts: support Palladium simulation

klin02:
To add extra Macros when pldm-build, use PLDM_EXTRA_MACRO="..." 
To add extra Args when make pldm-run/pldm-debug, use PLDM_EXTRA_ARGS="..." 
Use RELEASE_WITH_ASSERT to specify Palladium simulation without Software.


palladium.mk
```diff
@@ -0,0 +1,128 @@
+PLDM_TB_TOP  		 = tb_top
+PLDM_TOP_MODULE 	 = SimTop
```
poemonsense, line 2: This macro is already defined in Makefile, the SIM_TOP. 

should assign SIM_TOP to PLDM_TOP_MODULE

scripts/palladium/run_debug.tcl
```diff
@@ -0,0 +1,13 @@
+debug .
+host bjos_emu
```
poemonsense, line 2: I don't think we should expose the internal host name and other information to the user in difftest repo.

Probably we can wrap these scripts in Python and generate them with runtime arguments.

But we can first merge this PR and write the Python scripts in the next if you prefer

poemonsense (commented):
Please check again and remove duplicated variables in the makefiles.

The scripts for running palladium seems not suitable to be put into the git repo because everyone would have its own simulation arguments. For now, I'm good to merge this PR but we should think about this question in the next.

klin02:
Sure. I'll try to hide internal information.

palladium.mk
```diff
@@ -0,0 +1,127 @@
...
+IXCOM_FLAGS 	+= +dut+$(PLDM_TB_TOP)
+ifeq ($(RELEASE_WITH_ASSERT), 1)
+PLDM_CLOCK	 = clock_gen
+PLDM_CLOCK_DEF 	 = $(PLDM_SCRIPTS_DIR)/$(PLDM_CLOCK).xel
+PLDM_CLOCK_SRC 	 = $(PLDM_BUILD_DIR)/$(PLDM_CLOCK).sv
+IXCOM_FLAGS    	+= +dut+$(PLDM_CLOCK) $(PLDM_CLOCK_SRC)
+endif
+
+# Other Args
+AXIS_HOME 	 = /nfs/tools/Cadence/IXCOM22.04.s003/share/vxe
```
poemonsense, line 43: This should be removed

poemonsense (approved):

poemonsense:
Better to provide clear interface to specify the required arguments/paths. Only to hide internal paths but external users can still use this repo to run simulation

poemonsense:
The scripts may be added later. At that time we should provide a clear interface file which can be user defined for simulation. The only thing external users (and us) need to do is set the user defined paths in this file.

PR #250, opened 2024-01-15
Use DPIC-function to return deferred result

klin02:
Before this commit, we fetch simv_result at 5000 cycles. Only when the result is not zero, it works and stop simulation.

Now we use DPIC task to implement the return of simv_result when not zero, so only one DPIC task will be used to stop simulation. That will reduce the cost of fetch and get result return faster.

### DIFF
--- a/src/test/vsrc/vcs/DeferredControl.v
+++ b/src/test/vsrc/vcs/DeferredControl.v
@@ -7,8 +7,18 @@ module DeferredControl(
   output reg simv_result
 );
 
-import "DPI-C" function int simv_result_fetch();
 import "DPI-C" function void simv_nstep(int step);
+import "DPI-C" context function void set_deferred_result_scope();
+
+initial begin
+  set_deferred_result_scope();
+  simv_result = 1'b0;
+end
+
+export "DPI-C" function set_deferred_result;
+function void set_deferred_result();
+  simv_result = 1'b1;
+endfunction
 
 `ifdef CONFIG_DIFFTEST_NONBLOCK
 `ifdef PALLADIUM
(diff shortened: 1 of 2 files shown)


src/test/vsrc/vcs/DeferredControl.v
```diff
@@ -12,38 +12,28 @@ module DeferredControl(
...
-      end
-    end
-    else begin
-      fetch_timer <= fetch_timer + 64'b1;
-    end
-
-    if ((~simv_result) && (|step)) begin
-      simv_nstep(step);
-    end
+  if (~reset && (~simv_result) && (|step)) begin
```
poemonsense, line 35: Better to use `!` instead of `~`, `step != 0` instead of `|step` for logic operations.

src/test/vsrc/vcs/DeferredControl.v
```diff
@@ -12,38 +12,28 @@ module DeferredControl(
...
-      end
-    end
-    else begin
-      fetch_timer <= fetch_timer + 64'b1;
-    end
-
-    if ((~simv_result) && (|step)) begin
-      simv_nstep(step);
-    end
+  if ((!reset) && (!simv_result) && (step != 0)) begin
```
poemonsense, line 35: No need for the extra braces `()` but it's ok

poemonsense (dismissed):

poemonsense (approved):

klin02:
The DPIC-task is contained in GFIFO function，which unsupport output. I will check whether it works

klin02:
> The DPIC-task is contained in GFIFO function，which unsupport output. I will check whether it works
To return a value in GFIFO function, we can use SFIFO DPIC function to set it.

poemonsense (approved):

PR #251, opened 2024-01-15
Gen C++ and Verilog macros by GatewayConfig

klin02:
Currently all macros in Gateway can be decided by GatewayConfig, we move macros inside Config for better comprehension.

Difftest-related macros in Verilog can be configured in generated DifftestMacros.v, which should be included by all files use those macros.

Some redundant passing of config and return val of collect is removed. Some macros are renamed for similar format.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,7 @@ PLUGIN_CSRC_DIR = $(abspath ./src/test/csrc/plugin)
 PLUGIN_INC_DIR  = $(abspath $(PLUGIN_CSRC_DIR)/include)
 SIM_CXXFLAGS   += -I$(PLUGIN_INC_DIR)
 
+GEN_VSRC_DIR = $(BUILD_DIR)/generated-src
 VSRC_DIR   = $(abspath ./src/test/vsrc/common)
 SIM_VSRC = $(shell find $(VSRC_DIR) -name "*.v" -or -name "*.sv")
 
--- a/src/main/scala/Difftest.scala
+++ b/src/main/scala/Difftest.scala
@@ -279,7 +279,8 @@ trait DifftestModule[T <: DifftestBundle] {
 object DifftestModule {
   private val enabled = true
   private val instances = ListBuffer.empty[(DifftestBundle, String)]
-  private val macros = ListBuffer.empty[String]
+  private val cppMacros = ListBuffer.empty[String]
+  private val vMacros = ListBuffer.empty[String]
 
   def apply[T <: DifftestBundle](
     gen:      T,
@@ -313,13 +314,15 @@ object DifftestModule {
     difftest.step := 0.U
 
     val gateway = Gateway.collect()
-    macros ++= gateway._1
-    instances ++= gateway._2
-    difftest.step := gateway._3
+    cppMacros ++= gateway._1
+    vMacros ++= gateway._2
+    instances ++= gateway._3
+    difftest.step := gateway._4
 
     if (cppHeader.isDefined) {
       generateCppHeader(cpu, cppHeader.get)
     }
+    generateVeriogHeader()
 
     val timer = RegInit(0.U(64.W)).suggestName("timer")
     timer := timer + 1.U
@@ -343,7 +346,7 @@ object DifftestModule {
     difftestCpp += "#include <cstdint>"
     difftestCpp += ""
 
-    macros.foreach(m => difftestCpp += s"#define $m")
+    cppMacros.foreach(m => difftestCpp += s"#define $m")
     difftestCpp += ""
 
     val cpu_s = cpu.replace("-", "_").replace(" ", "").toUpperCase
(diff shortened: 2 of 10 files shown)


src/main/scala/Difftest.scala
```diff
@@ -415,6 +418,15 @@ object DifftestModule {
     val outputFile = outputDir + "/diffstate.h"
     Files.write(Paths.get(outputFile), difftestCpp.mkString("\n").getBytes(StandardCharsets.UTF_8))
   }
+
+  def generateVeriogHeader(): Unit = {
+    val difftestVeriog = ListBuffer.empty[String]
+    vMacros.foreach(m => difftestVeriog += s"`define $m")
+    val outputDir = sys.env("NOOP_HOME") + "/build/rtl"
```
poemonsense, line 425: Generate files must be put in `generated-src` only to avoid confusions. `build/rtl` is only for chisel design files.

src/test/vsrc/vcs/top.v
```diff
@@ -13,11 +13,8 @@
 *
 * See the Mulan PSL v2 for more details.
 ***************************************************************************************/
-`ifdef TB_DPIC_NONBLOCK
-  `define TB_DEFERRED_RESULT
-`endif
-`define STEP_WIDTH 8
 
+`include "DifftestMacros.v"
```
poemonsense, line 17: what if difftest is not enabled? This file will not exist

src/main/scala/Gateway.scala
```diff
@@ -32,16 +32,38 @@ case class GatewayConfig(
                         replaySize     : Int     = 256,
                         diffStateSelect: Boolean = false,
                         isBatch        : Boolean = false,
-                        batchSize      : Int     = 32
+                        batchSize      : Int     = 32,
+                        isNonBlock     : Boolean = false
```
poemonsense, line 36: Always add a `,` after the arguments. Then adding arguments will not modify the previous line. `isNonBlock: Boolean = false,`

src/main/scala/Difftest.scala
poemonsense: Make the output directory a function shared by generateVeriogHeader and generateCppHeader

src/main/scala/Gateway.scala
```diff
@@ -32,16 +32,38 @@ case class GatewayConfig(
...
+    macros += s"CONFIG_DIFFTEST_BUFLEN ${dutBufLen}"
+    if (isBatch) macros ++= Seq("CONFIG_DIFFTEST_BATCH", s"DIFFTEST_BATCH_SIZE ${batchSize}")
+    if (isSquash) macros += "CONFIG_DIFFTEST_SQUASH"
+    if (squashReplay) macros += "CONFIG_DIFFTEST_SQUASH_REPLAY"
+    if (hasDeferredResult) macros += "CONFIG_DIFFTEST_DEFERRED_RESULT"
+    macros.toSeq
+  }
+  def vMacros: Seq[String] = {
+    val macros = ListBuffer.empty[String]
+    macros += s"TB_DIFF_STEP_WIDTH ${stepWidth}"
```
poemonsense, line 62: why not also use the `CONFIG_DIFFTEST_*` names?

One of the good naming practice is to avoid abbreviations if acceptable.

poemonsense (approved):

src/test/vsrc/vcs/top.v
klin02: It seems DifftestModule.finish() will always be called by SimTop, and generate-src will always be include in Difftest/Makefile

src/main/scala/Difftest.scala
klin02: Maybe we can move the function outside of DifftestModule so that it can also be shared by difftest-dpic.c/h ?

poemonsense: Sure.

src/test/vsrc/vcs/top.v
poemonsense: Look good to me

poemonsense:
CI fails because NutShell changes the RTL path. I'll fix it later

Makefile
```diff
@@ -64,7 +64,8 @@ PLUGIN_CSRC_DIR = $(abspath ./src/test/csrc/plugin)
 PLUGIN_INC_DIR  = $(abspath $(PLUGIN_CSRC_DIR)/include)
 SIM_CXXFLAGS   += -I$(PLUGIN_INC_DIR)
 
-VSRC_DIR   = $(abspath ./src/test/vsrc/common)
+GEN_VSRC_DIR = $(BUILD_DIR)/generated-src
+VSRC_DIR   = $(abspath ./src/test/vsrc/common) $(GEN_VSRC_DIR)
```
poemonsense, line 68: No need to add  $(GEN_VSRC_DIR) to VSRC_DIR?

VCS and verilator will search for the missing modules in include paths

poemonsense (dismissed):

Makefile
klin02: Sure, I will delete it. Now I have  trouble using cross-directory include on Palladium，and I am asking Palladium engineer for help.

poemonsense: what about passing all verilog files, including the generated ones, to palladium?

poemonsense (approved):

PR #261, opened 2024-01-20
Allow LogPerfControl to be reused if necessary

poemonsense:
Since https://github.com/chipsalliance/chisel/pull/3753, the users are able to check whether a Data is visible in the current context. To avoid a large number of duplicated LogPerfControl modules, we allow it to be reused when some previous instance is still visible to the current context.

Call `LogPerfControl.reuse(DataMirror.isVisible)` instead of pure `apply()` to allow the reuse.

However, this is still a temporary workaround. In the future, we need some clean methods to extract the simulation environment and access these signals during simulation, possibly via XMR.

### DIFF
--- a/src/main/scala/common/LogPerfControl.scala
+++ b/src/main/scala/common/LogPerfControl.scala
@@ -25,7 +25,7 @@ class LogPerfControl extends Bundle {
   val dump = Bool()
 }
 
-class LogPerfHelper extends BlackBox with HasBlackBoxInline {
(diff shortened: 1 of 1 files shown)


poemonsense:
To maintain the backward compatibility with older Chisel versions, we cannot include `DataMirror.isVisible` here in difftest. Update-to-date Chisel users are expected to passing this checker function to difftest explicitly.

PR #268, opened 2024-01-28
Gateway: wrap results in a case class

poemonsense:
We wrap the gateway return values in a case class for better code readability and code structure.

The results of the gateway consist of:
* C++ macros
* Verilog macros
* Extra difftest bundles
* DiffTest step


klin02 (approved):
Great! Readability improve lot!
I am also working in seperate Gateway component now for better code structure.

PR #271, opened 2024-01-28
Gateway: seperate functional component

klin02:
* We seperate gateway mixed logic to PreProcess, Squash, Batch, GatewaySink and ZoneControl
* We add zone-related logic. DUT will write to different zones of State_Buffer at different clock. And REF will also read different zone at each nstep. It will help with async Read and Write. Also used when step at next cycle may be lag behind transfer.
* Each zone's len will be batchSize when isBatch, default is 1.
* Transfer Module such as DPIC should extends GatewaySink for same IO interface. And batch will has specific Interface, extends GatewayBatchSink.

### DIFF
--- a/src/main/scala/Batch.scala
+++ b/src/main/scala/Batch.scala
@@ -1,0 +1,57 @@
+/***************************************************************************************
+ * Copyright (c) 2020-2024 Institute of Computing Technology, Chinese Academy of Sciences
+ *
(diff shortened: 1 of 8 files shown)


klin02:
It seems that to use scalaFmt, we can only use branch of this repo, so bot can commit and push

PR #272, opened 2024-01-28
Gateway: seperate functional component

klin02:
* We seperate gateway mixed logic to PreProcess, Squash, Batch, GatewaySink and ZoneControl
* We add zone-related logic. DUT will write to different zones of State_Buffer at different clock. And REF will also read different zone at each nstep. It will help with async Read and Write. Also used when step at next cycle may be lag behind transfer.
* Each zone's len will be batchSize when isBatch, default is 1.
* Transfer Module such as DPIC should declare GatewayBundle for same IO interface. And batch will has specific Interface, declare GatewayBatchBundle.

### DIFF
--- a/src/main/scala/DPIC.scala
+++ b/src/main/scala/DPIC.scala
@@ -21,7 +21,7 @@ import chisel3.reflect.DataMirror
 import chisel3.util._
 import difftest.DifftestModule.streamToFile
 import difftest._
-import difftest.gateway.{GatewayBundle, GatewayConfig}
+import difftest.gateway.{GatewayBatchBundle, GatewayBundle, GatewayConfig}
 
 import scala.collection.mutable.ListBuffer
 
@@ -32,7 +32,8 @@ class DPIC[T <: DifftestBundle](gen: T, config: GatewayConfig)
   val clock = IO(Input(Clock()))
   val enable = IO(Input(Bool()))
   val io = IO(Input(gen))
-  val dut_pos = Option.when(config.hasDutPos)(IO(Input(UInt(config.dutPosWidth.W))))
+  val dut_zone = Option.when(config.hasDutZone)(IO(Input(UInt(config.dutZoneWidth.W))))
+  val dut_index = Option.when(config.isBatch)(IO(Input(UInt(log2Ceil(config.batchSize).W))))
 
   def getDirectionString(data: Data): String = {
     if (DataMirror.directionOf(data) == ActualDirection.Input) "input " else "output"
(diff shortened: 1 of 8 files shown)


klin02:
It seems latest commit "make format" will not trigger actions?

src/main/scala/DPIC.scala
```diff
@@ -105,8 +110,9 @@ class DPIC[T <: DifftestBundle](gen: T, config: GatewayConfig)
        |  ${dpicFuncArgs.flatten.map(arg => getDPICArgString(arg._1, arg._2, true)).mkString(",\n  ")}
        |)""".stripMargin
   val dpicFunc: String = {
-    val dut_pos = if (config.hasDutPos) "dut_pos" else "0"
-    val packet = s"DUT_BUF(io_coreid,$dut_pos)->${gen.desiredCppName}"
+    val dut_zone = if (config.hasDutZone) "dut_zone" else "0"
+    val dut_pos = if (config.isBatch) "dut_pos" else "0"
+    val packet = s"DUT_BUF(io_coreid,$dut_zone,$dut_pos)->${gen.desiredCppName}"
```
poemonsense, line 115: whitespace after `,`

src/main/scala/DPIC.scala
```diff
@@ -205,20 +228,30 @@ object DPIC {
...
         |  }
-        |  inline DiffTestState* get(int pos) {
-        |    return buffer+pos;
+        |  inline DiffTestState* get(int zone, int pos) {
+        |    return buffer[zone]+pos;
         |  }
         |  inline DiffTestState* next() {
-        |    DiffTestState* ret = buffer+read_ptr;
-        |    read_ptr = (read_ptr + 1) % CONFIG_DIFFTEST_BUFLEN;
+        |    DiffTestState* ret = buffer[zone_ptr]+read_ptr;
```
poemonsense, line 243: whitespaces around `+`

src/main/scala/DPIC.scala
```diff
@@ -236,7 +269,7 @@ object DPIC {
     interfaceCpp +=
       s"""
          |DiffStateBuffer** diffstate_buffer = nullptr;
-         |#define DUT_BUF(core_id,pos) (diffstate_buffer[core_id]->get(pos))
+         |#define DUT_BUF(core_id,zone,pos) (diffstate_buffer[core_id]->get(zone,pos))
```
poemonsense, line 272: whitespace after `,`

src/main/scala/Difftest.scala
```diff
@@ -390,8 +390,9 @@ object DifftestModule {
          |class DiffStateBuffer {
          |public:
          |  virtual ~DiffStateBuffer() {}
-         |  virtual DiffTestState* get(int pos) = 0;
+         |  virtual DiffTestState* get(int zone, int pos) = 0;
```
poemonsense, line 393: any other variable naming for `pos`? `zone` and `pos` are difficult to distinguish their meanings.

src/main/scala/Gateway.scala
```diff
@@ -117,15 +120,133 @@ object Gateway {
...
   val in_pack = WireInit(0.U.asTypeOf(MixedVec(signals.map(gen => UInt(gen.getWidth.W)))))
   for ((data, id) <- in_pack.zipWithIndex) {
     DifftestWiring.addSink(data, s"gateway_$id")
     in(id) := data.asTypeOf(in(id).cloneType)
   }
 
-  val share_wbint = WireInit(in)
-  if (config.hasDutPos || config.isSquash) {
+  val preprocessed = if (config.needPreprocess) {
+    val preprocess = Module(new Preprocess(in.toSeq.map(_.cloneType), config))
```
poemonsense, line 132: Let `Preprocess` have a input of `in` and `config`. Then wrap the cloneType into the apply as well as the `preprocess.in := in`. Output `preprocess.out` in the apply function.

Similar for `Squash` and `Batch`.

This change can simplify the code and improve readability

poemonsense (approved):

PR #273, opened 2024-01-29
CI: modify sed for scalaFmt

klin02:
Indentation style of GatewayConfig param is changed by scalaFmt. We modify sed to adapting that.


klin02:
After this commit, CI works as predicted.

poemonsense (approved):

PR #281, opened 2024-02-02
Batch: pack DPIC to reduce sync times

klin02:
We pack DPIC of DifftestBundles on Hardware and unpack it in Software. So tens of DPIC called times will be reduced to one, which reduce sync times in platform such as palladium.

It can speed up full-difftest of XiangShan from 0.23MHz to 0.29MHz with GlobalEnable,Squash and Gfifo(Non-block) in Palladium. Without Gfifo, it can speed up from 0.025 MHz to 0.11MHz.

Difftest cycle delayed for DifftestBundles to collect valid bundles in bundleNum cycles, which cause delayed comparision result.

BatchInterval and BatchFinish are appended to infoVec to seperate DPIC packed for different Cycle. Other elements contains BundleType and BundleLen to help unpack DataVec.

Now we set maximum data byte packed as 3900, and maximum info byte packed as 90, because Palladium limits total bytes of gfifo param should less than 4000.

We add `--max-num-width` verilator args to increase bit width limits. See related issue https://github.com/verilator/verilator/issues/2082

For misalignment of struct member, we add packed attribute to struct declaration so all member will be aligned by byte, which helps use memcpy for DPIC.

Note: When run full-difftest in XiangShan with Batch mode, it is possible to contains only one cycle DPICs each time for larger possible maximum bytes each cycle. So diffStateSelect should also be open for VCS and Palladium

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -320,7 +320,7 @@ jobs:
             ./build/simv +workload=./ready-to-run/microbench.bin +e=0 +no-diff +max-cycles=100000
             ./build/simv +workload=./ready-to-run/microbench.bin +e=0 +diff=./ready-to-run/riscv64-nemu-interpreter-so
 
-      - name: Verilator Build with VCS Top (with Squash Batch and Global Enable)
+      - name: Verilator Build with VCS Top (with Squash Select Batch and Global Enable)
         run: |
             cd $GITHUB_WORKSPACE/../xs-env
             source ./env.sh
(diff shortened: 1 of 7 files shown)


poemonsense (commented):
Overall, there seems some duplicated code, and we need to remove them.

We need to find a clean way to implement DPI-C batch. It looks like another implementation for DPI-C. The original DPI-C is actually single-packet DPI-C. This new one is batch DPI-C (using DPI-C to forward the batch data to difftest). Thus, there could be several processing stages.

For data process, we may use DifftestBundle (the original implementation), UInt generated by `asUInt` (current implementation), and batch data (this newly proposed one). This stages look like what `Gateway` is expected to work.

For data collection from the hardware, we may use DifftestBundle (the current DPI-C), or a large UInt (this newly proposed one). In the future, we may also provide more ways, such as Palladium internal API, FPGA PCIe, etc. This stage looks like the current what the `GatewaySink` should work.

So, we still need to re-think the passes in Difftest. Some of them are optimization passes only (algorithmic transformations only at the RTL level, independent of the underlying platform). Some other of them are platform-specific communication methods, such as DPI-C, Palladium internal API, ....And these methods may accept different types of Data from the previous stage, such as Seq[Bundle], Seq[UInt], or, in this newly proposed optimization pass, a very wide UInt.

We need to be clear 1) the input/output of each stage and each optimization pass; 2) what the stage or optimization pass is for; 3) the relationship between the passes/stages, such as inheritance, for better code structure.

klin02:
For Batch, we add abstract class DPICBase for common methods and template. DPIC/DPICBatch only need to override little logic to reuse it.
For DPIC, we seperate io and control, because both Single/Batch will use same control interface, but different io interface. We don't need to declare extra bundle for that.

poemonsense (approved):
Very good work.

PR #286, opened 2024-02-07
palladium: add scripts

klin02:
Now we release related palladium scripts and user should set PLDM_HOST for internal information.
We use wildcard for boards occupation in compilerOptions.qel, it can automatically choose approprate boards count. User can also set it manully like {boards 0+1.1+1.2}, which means occupy 1 board and 2 extra domains. It can be estimated by gate count in xe.msg.

### DIFF
--- a/scripts/palladium/argConfigs.qel
+++ b/scripts/palladium/argConfigs.qel
@@ -1,0 +1,4 @@
+* $test$plusargs TB_IMPORT
+* $value$plusargs TB_IMPORT
+* $finish TB_IMPORT
+* $random TB_IMPORT
--- a/scripts/palladium/run.tcl
+++ b/scripts/palladium/run.tcl
@@ -1,0 +1,6 @@
+debug .
+host $env(PLDM_HOST)
+xc on -zt0 -xt0 -tbrun
+date
+run -swap
+run
--- a/palladium.mk
+++ b/palladium.mk
@@ -1,6 +1,7 @@
 PLDM_TB_TOP  		 = tb_top
 PLDM_TOP_MODULE 	 = $(SIM_TOP)
 
+PLDM_SCRIPTS_DIR 	 = $(abspath ./scripts/palladium)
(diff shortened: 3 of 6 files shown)


scripts/palladium/argConfigs.qel
```diff
@@ -0,0 +1,4 @@
+* $test$plusargs TB_IMPORT
+* $value$plusargs TB_IMPORT
+* $finish TB_IMPORT
+* $random TB_IMPORT
```
poemonsense, line 4: This red mark shown by GitHub means there's no trailing `\n` (no newline) at the end of this file.

Leave a `\n` after the last line (adding a blank line in vscode, or just save the file in `vim` should work).

scripts/palladium/run.tcl
```diff
@@ -0,0 +1,6 @@
+debug .
```
poemonsense, line 1: I'm still worried about whether we should provide the simulation scripts. Will they change frequently? Are they dependent of the underlying palladium platform?

I'm not familiar with palladium. If the answers are not, I'm good with the changes.

poemonsense (approved):
Overall, look good to me. Feel free to merge it.

scripts/palladium/argConfigs.qel
poemonsense: The red mark is not shown in the review window. It will be shown in the "Files changed" tab.

scripts/palladium/run.tcl
klin02: Based on our usage, the emulation script does not need to be changed and independent of underlying palladium platform version.

Take run.tcl as example:
`xc on -zt0 -xt0 -tbrun` means Hardware will set X and Z to 0, and has behaviour delay, which means block Hardware until Software func (except Gfifo Declared) called this cycle are done. 
`run -swap` means read memory from file. Which support we load workload by fread.
`run` starts simulation.

poemonsense (approved):

PR #288, opened 2024-02-12
Squash replay: optimized by NEMU store log

klin02:
Previous we use naive memcpy for NEMU memory snapshot. Now we complete more efficient memory snapshot and recovery with NEMU store log.

Call ref_store_log_reset and set_store_log, then we can record origin data of every pmem_write for this execution. When REF and DUT comparation trigger error, call ref_store_log_restore, then memory will recover in reverse order.

ENABLE_STORE_LOG macro added to config.


klin02:
This change should be supported by NEMU PR https://github.com/OpenXiangShan/NEMU/pull/265.
ready-to-run should be also updated.

klin02:
When bump to XiangShan with this commit, we should bump ready-to-run at the same time.

poemonsense (approved):

PR #289, opened 2024-02-14
Enable perfcnt for difftest

klin02:
We add perfcnt to evalute the effects of different optimization of Gateway. Now we cnt Dpic calls, bytes, run time and failed cycle. Speed and Gates will be reported by Palladium.


src/test/csrc/difftest/perf.h
```diff
@@ -0,0 +1,40 @@
+/***************************************************************************************
+* Copyright (c) 2020-2024 Institute of Computing Technology, Chinese Academy of Sciences
+* Copyright (c) 2020-2021 Peng Cheng Laboratory
```
poemonsense, line 3: remove pcl

src/test/csrc/difftest/perf.cpp
```diff
@@ -0,0 +1,57 @@
+/***************************************************************************************
+* Copyright (c) 2020-2024 Institute of Computing Technology, Chinese Academy of Sciences
+* Copyright (c) 2020-2021 Peng Cheng Laboratory
```
poemonsense, line 3: remove pcl

src/main/scala/DPIC.scala
```diff
@@ -339,6 +350,15 @@ object DPIC {
         |  }
         |};
         |""".stripMargin
+    interfaceCpp +=
+      s"""
+         |#ifdef CONFIG_DIFFTEST_PERFCNT
+         |enum DIFFSTATE_PERF {
+         |  ${(interfaces.map(_._1.replace("v_difftest_", "perf_")) ++ Seq("DIFFSTATE_PERF_NUM")).mkString(",\n  ")}
```
poemonsense, line 357: `replace("v_difftest_", "perf_")` is duplicated with line 85. Replace them with a func def?

src/main/scala/DPIC.scala
```diff
@@ -371,6 +391,28 @@ object DPIC {
...
+         |  for (int i = 0; i < DIFFSTATE_PERF_NUM; i++) {
+         |    long long calls_mean = dpic_calls[i] * 1000 / msec;
+         |    long long bytes_mean = dpic_bytes[i] * 1000 / msec;
+         |    total_calls += dpic_calls[i];
+         |    total_bytes += dpic_bytes[i];
+         |    total_calls_mean += calls_mean;
+         |    total_bytes_mean += bytes_mean;
+         |    printf(\"%20s %15lld %15lld/s %15lldB %15lldB/s\\n\", dpic_name[i], dpic_calls[i], calls_mean, dpic_bytes[i], bytes_mean);
+         |  }
+         |  printf(\"%20s %15lld %15lld/s %15lldB %15lldB/s\\n\", \"TOTAL\", total_calls, total_calls_mean, total_bytes, total_bytes_mean);
```
poemonsense, line 412: These `printf("%20s %15lld %15lld/s\n"` seems duplicated as well in perf.cpp. If duplicated, move them into perf.cpp and reuse the function here?

src/test/csrc/difftest/perf.cpp
```diff
@@ -0,0 +1,42 @@
+#ifdef CONFIG_DIFFTEST_PERFCNT
```
poemonsense, line 1: License is required.  See https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/common/coverage.cpp

src/test/csrc/difftest/perf.h
```diff
@@ -0,0 +1,29 @@
+#ifndef __PERF_H__
+#define __PERF_H__
+
+#ifdef CONFIG_DIFFTEST_PERFCNT
+inline void difftest_perfcnt_print(const char* name, long long calls, long long bytes, long long msec) {
```
poemonsense, line 5: Functions in header files should be static to avoid conflicts.

poemonsense (commented):
It looks like the perf files should be under common instead of difftest. difftest directory is for checking the dut behaviors against ref. common directory is for the common simulation. Even if difftest (checking results) is disabled, it should also have these perf_* calls, such as perf_difftest_ram_read, ....?

(Since we are putting the simulation environment in this difftest repo, these namings seem weird. This is a long story.)

If this change is too much, it's ok to merge it now. We will refactor it in the future.

poemonsense (approved):

PR #290, opened 2024-02-19
vcs: add support for --max-instrs option

xiaokamikami:
Add a per-instruction counter to exit, primarily for checkpoint testing

### DIFF
--- a/src/test/csrc/vcs/vcs_main.cpp
+++ b/src/test/csrc/vcs/vcs_main.cpp
@@ -30,6 +30,7 @@ static bool has_reset = false;
 static char bin_file[256] = "ram.bin";
 static char *flash_bin_file = NULL;
 static bool enable_difftest = true;
+static uint64_t max_instrs = 0;
 
 extern "C" void set_bin_file(char *s) {
   printf("ram image:%s\n",s);
@@ -40,6 +41,10 @@ extern "C" void set_flash_bin(char *s) {
   printf("flash image:%s\n",s);
   flash_bin_file = (char *)malloc(256);
   strcpy(flash_bin_file, s);
+}
+
+extern "C" void set_max_instrs(uint64_t mc) {
+  max_instrs = mc;
 }
 
 extern const char *difftest_ref_so;
@@ -84,6 +89,14 @@ extern "C" int simv_step() {
         eprintf(ANSI_COLOR_RED "Unknown trap code: %d\n" ANSI_COLOR_RESET, trapCode);
     }
     return trapCode + 1;
+  }
+
+  if (max_instrs != 0) { // 0 for no limit
+    auto trap = difftest[0]->get_trap_event();
+    if(max_instrs < trap->instrCnt) {
+      eprintf(ANSI_COLOR_GREEN "EXCEEDED MAX INSTR: %ld\n" ANSI_COLOR_RESET,max_instrs);
+      return 3;// STATE_LIMIT_EXCEEDED
+    }
   }
 
   if (enable_difftest) {
--- a/src/test/vsrc/vcs/top.v
+++ b/src/test/vsrc/vcs/top.v
@@ -23,6 +23,7 @@ import "DPI-C" function void set_flash_bin(string bin);
 import "DPI-C" function void set_diff_ref_so(string diff_so);
 import "DPI-C" function void set_no_diff();
 import "DPI-C" function void simv_init();
+import "DPI-C" function void set_max_instrs(longint mc);
 `ifndef CONFIG_DIFFTEST_DEFERRED_RESULT
 import "DPI-C" function int simv_nstep(int step);
 `endif // CONFIG_DIFFTEST_DEFERRED_RESULT
@@ -55,6 +56,7 @@ string bin_file;
 string flash_bin_file;
 string wave_type;
 string diff_ref_so;
+reg [63:0] max_instrs;
 reg [63:0] max_cycles;
 
 initial begin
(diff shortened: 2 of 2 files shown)


src/test/csrc/difftest/difftest.cpp
```diff
@@ -87,6 +87,20 @@ int difftest_step() {
...
+uint64_t difftest_commit_sum(char core_id) {
+  if (core_id < NUM_CORES)
+    return difftest[core_id]->get_instr_sum();
+  else 
+    return 0;
+}
+
+void difftest_commit_clean() {
+  for (uint32_t core_id = 0; core_id < NUM_CORES; core_id++)
+  {
```
poemonsense, line 99: Follow the same code style as existing code (to put `{` at the same line of `for`)

poemonsense (commented):
let @klin02 have a review and test

src/test/csrc/difftest/difftest.cpp
xiaokamikami: OK

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -42,6 +43,10 @@ extern "C" void set_flash_bin(char *s) {
   strcpy(flash_bin_file, s);
 }
 
+extern "C" void set_max_instrs(long mc) {
```
klin02, line 46: It seems type `long` mismatchs DPI-C declare? Better use same type and more clear name then `mc`~

klin02:
Better add exit Info like https://github.com/OpenXiangShan/difftest/blob/0330732fcdd22c7209866a3ae2213c8b29a4605b/src/test/vsrc/vcs/top.v#L199

src/test/csrc/difftest/difftest.cpp
```diff
@@ -87,6 +87,19 @@ int difftest_step() {
...
 }
 
+uint64_t difftest_commit_sum(char core_id) {
+  if (core_id < NUM_CORES)
+    return difftest[core_id]->get_instr_sum();
+  else 
+    return 0;
+}
+
+void difftest_commit_clean() {
```
klin02, line 97: When will difftest_commit_clean be used? 
By the way, better use same code style as other difftest_xxx Func.
for (int i = 0; i < NUM_CORES; i++) ...

Same for other new function.

klin02:
Does retire_instr_cntsum have any use other than controlling instruction overflow exit? Can trap->instrcnt be used as same role like emu?
https://github.com/OpenXiangShan/difftest/blob/0330732fcdd22c7209866a3ae2213c8b29a4605b/src/test/csrc/verilator/emu.cpp#L698
If some unused function is necessary for checkpoint. Maybe you can explain it in this PR?

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -25,11 +25,13 @@
 #ifdef CONFIG_DIFFTEST_DEFERRED_RESULT
 #include "svdpi.h"
 #endif // CONFIG_DIFFTEST_DEFERRED_RESULT
+#define EXIT_MAX_INSTR 0xff
 
 static bool has_reset = false;
 static char bin_file[256] = "ram.bin";
 static char *flash_bin_file = NULL;
 static bool enable_difftest = true;
+static int max_instrs = 0;
```
klin02, line 34: use uint64_t? Line 47 arg is uint64_t, and trap->instrCnt is also uint64_t

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -86,6 +92,14 @@ extern "C" int simv_step() {
     return trapCode + 1;
   }
 
+  if (max_instrs != 0) { // 0 for no limit
+    auto trap = difftest[0]->get_trap_event();
+    if(max_instrs < trap->instrCnt) {
+      eprintf(ANSI_COLOR_GREEN "checkpoint reach the operating limit exit\n" ANSI_COLOR_RESET);
```
klin02, line 98: Seems --max-instrs is independent to checkpoint. Simply `EXCEEDED MAX INSTR: %d` is enough

klin02 (changes requested):

poemonsense:
In Linux, the return code of a process indicates the state of the process. See https://www.agileconnection.com/article/overview-linux-exit-codes

Return code 0 means the program exits normally. For max-instrs, the simulation should return 0 as well. This is also the normal behavior in the Verilator simulation as shown in https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/common/main.cpp#L45 and https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/verilator/emu.h#L138

Please do not use 0xff as the return code

xiaokamikami:
> In Linux, the return code of a process indicates the state of the process. See https://www.agileconnection.com/article/overview-linux-exit-codes
> 
> Return code 0 means the program exits normally. For max-instrs, the simulation should return 0 as well. This is also the normal behavior in the Verilator simulation as shown in https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/common/main.cpp#L45 and https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/verilator/emu.h#L138
> 
> Please do not use 0xff as the return code

In the next version I need to determine if it's max-instrs and run a separate handler

poemonsense:
> > In Linux, the return code of a process indicates the state of the process. See https://www.agileconnection.com/article/overview-linux-exit-codes
> > Return code 0 means the program exits normally. For max-instrs, the simulation should return 0 as well. This is also the normal behavior in the Verilator simulation as shown in https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/common/main.cpp#L45 and https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/verilator/emu.h#L138
> > Please do not use 0xff as the return code
> 
> In the next version I need to determine if it's max-instrs and run a separate handler

OK. It's only for internal usage

src/test/csrc/vcs/vcs_main.cpp
xiaokamikami: fix

src/test/csrc/vcs/vcs_main.cpp
xiaokamikami: fix

klin02 (approved):

PR #291, opened 2024-02-20
Decompress: support zstd algorithm

xyyy1420:


Makefile
```diff
@@ -134,6 +134,14 @@ else
 SIM_LDFLAGS  += -lz
 endif
 
+# ZSTD image support
+IMAGE_ZSTD_COMPRESS ?= 1
+ifeq ($(IMAGE_ZSTD_COMPRESS),0)
```
poemonsense, line 139: If by default it is enabled, then we can use `NO_ZSTD_COMPRESSION=0` to disable it. No need to use `IMAGE_ZSTD_COMPRESS ` with default 1.

src/test/csrc/common/compress.cpp
```diff
@@ -126,6 +134,122 @@ long readFromGz(void* ptr, const char *file_name, long buf_size, uint8_t load_ty
...
+  if (ZSTD_isError(decompress_result) || output_buffer.pos!=0) {
+    printf("Decompress failed: %s\n", ZSTD_getErrorName(decompress_result));
+    printf("Binary size larger than memory\n");
+    ZSTD_freeDStream(dstream);
+    delete [] compress_file_buffer;
+    delete [] temp_page;
+    return -1;
+  }
+
+  // printf("Read %lu bytes from gz stream in total\n", curr_size);
```
poemonsense, line 241: remove unused printf

src/test/csrc/common/compress.cpp
```diff
@@ -126,6 +134,122 @@ long readFromGz(void* ptr, const char *file_name, long buf_size, uint8_t load_ty
...
+  uint8_t *compress_file_buffer = NULL;
+  size_t compress_file_buffer_size = 0;
+
+  uint64_t curr_size = 0;
+  const uint32_t chunk_size = 16384;
+
+  // Only load from RAM need check
+  if (load_type == LOAD_RAM && (buf_size % chunk_size) != 0) {
+    printf("buf_size must be divisible by chunk_size\n");
+    assert(0);
```
poemonsense, line 152: return -1?

src/test/csrc/common/ram.cpp
```diff
@@ -274,6 +274,11 @@ MmapMemory::MmapMemory(const char *image, uint64_t n_bytes) : SimMemory(n_bytes)
     img_size = readFromGz(ram, image, memory_size, LOAD_RAM);
     assert(img_size >= 0);
   }
+  else if(isZstdFile(image)) {
```
poemonsense, line 277: It looks like we made a mistake not to wrap these `if` in macros.

It seems that, only when NO_ZSTD_COMPRESSION is not defined, we should check whether they are zstd files?

poemonsense (commented):
some minor changes

poemonsense:
CI fails due to missing zstd libraries. Is it installed by default in ubuntu 20.04? If not, we need to add it to xs-env/setup-tools.sh or disable it by default

src/test/csrc/common/ram.cpp
poemonsense: The current implementation is OK. We can change this behavior along with `.gz` together in the future.

xyyy1420: If we wrap these `if` in macros, when NO_ZSTD_COMPRESSION and  NO_GZ_COMPRESSION are not defined, the compressed data will be read into `memory`, as i understand, that would be a mistake?

src/test/csrc/common/compress.cpp
```diff
@@ -126,6 +134,122 @@ long readFromGz(void* ptr, const char *file_name, long buf_size, uint8_t load_ty
...
+
+  // printf("Read %lu bytes from gz stream in total\n", curr_size);
+
+  ZSTD_freeDStream(dstream);
+  delete [] compress_file_buffer;
+  delete [] temp_page;
+
+  return curr_size;
+#else
+  return 0;
```
xyyy1420, line 249: Maybe we can `return -1` when NO_ZSTD_COMPRESSION and NO_GZ_COMPRESSION are not defined?

src/test/csrc/common/ram.cpp
poemonsense: Yes, gz/zstd files will be read as normal files and may cause errors. I think you are right.

src/test/csrc/common/compress.cpp
poemonsense: I think it would be better to allow the user to specify any file as input. If compression is not enabled and the user inputs a gz file, then it is read as a normal file but raise a warning to tell the user there may be something wrong?

I think all are OK for us. At lease normal users would not pass a gz file as a normal input. So the current implementation is ok, to raise errors when input gz without compression

xyyy1420:
> CI fails due to missing zstd libraries. Is it installed by default in ubuntu 20.04? If not, we need to add it to xs-env/setup-tools.sh or disable it by default

I create a pr for add it to xs-env/setup-tools.sh. https://github.com/OpenXiangShan/xs-env/pull/41

Makefile
xyyy1420: I don't quite get it, should I keep the `IMAGE_ZSTD_COMPRESS`?
If i remove it, when we use `NO_ZSTD_COMPRESSION=0` to disable decompress with zstd, `-lzstd` will be useless?

src/test/csrc/common/compress.cpp
xyyy1420: Okay, I'll keep it.

Makefile
poemonsense:
```Makefile
ifneq ($(NO_ZSTD_COMPRESSION),)
SIM_CXXFLAGS += -DNO_ZSTD_COMPRESSION
else
SIM_LDFLAGS  += -lzstd
endif
```

poemonsense: But the current implementation is also good. No need to change.

poemonsense (approved):

PR #293, opened 2024-02-23
vcs: add support for --workload-list option

xiaokamikami:
When you use --workload-list to set up a list of checkpoints or workload, you run through all the workloadin the list in a single launch; Using this feature requires need add WORKLOAD_SWITCH=1 at compile time and enable DPIC functionality
The format of the list should be，Use Spaces as separators
`work-path program-run-limit`


src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -162,10 +174,68 @@ extern "C" int simv_nstep(uint8_t step) {
...
+  fclose(fp);
+  difftest_finish();
+#ifdef CONFIG_DIFFTEST_DEFERRED_RESULT
+  simv_result = 0;
+#endif
+  return 0;
+}
+
+static int checkpoint_reset (char * file_name) {
+	int line_len = strlen(file_name);
```
poemonsense, line 223: please format the code like the others

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -162,10 +174,68 @@ extern "C" int simv_nstep(uint8_t step) {
...
+	if ('\n' == file_name[line_len - 1]) {
+		file_name[line_len - 1] = '\0';
+	}
+  char * str1 = strrchr(file_name, '/');
+  if (str1 != NULL)
+    str1 ++;
+  else 
+    return 1;
+
+  if (sscanf(str1,"_%d_0.%ld_.gz", &checkpoin_idx, &max_instrs) != 2) {
```
poemonsense, line 234: please format the code carefully. E.g., whitespace after `,`

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -58,6 +64,12 @@ extern "C" void set_diff_ref_so(char *s) {
   difftest_ref_so = buf;
 }
 
+extern "C" void difftest_checkpoint_list (char * path) {
+  checkpoint_list_path = (char *)malloc(256);
+  strcpy(checkpoint_list_path,path);
```
poemonsense, line 69: whitespace after ","

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -58,6 +64,12 @@ extern "C" void set_diff_ref_so(char *s) {
   difftest_ref_so = buf;
 }
 
+extern "C" void difftest_checkpoint_list (char * path) {
```
poemonsense, line 67: avoid whitespace between function name and arguments

poemonsense (commented):
1. This function seems not related to checkpoints. For all workloads, the list should work well. If so, then we should name it like `workload-list` instead of `ckpt-list`.

2. Probably we need a CI for this test? The list could be, like, multiple microbench.bin lines.

Let @klin02 have a review and help the CI things

xiaokamikami:
> 1. This function seems not related to checkpoints. For all workloads, the list should work well. If so, then we should name it like `workload-list` instead of `ckpt-list`.
> 2. Probably we need a CI for this test? The list could be, like, multiple microbench.bin lines.
> 
> Let @klin02 have a review and help the CI things

It's theoretically possible to support different workload lists, but for now I'm only supporting our new checkpoint resolution

poemonsense:
Also, we should create a strict C++ formatter to avoid PRs that bring in inconsistent C++ code style into the repo. @klin02 If you have time, please help find some C++ formatter that is suitable for us. I know Verilator is using the clang formatter for C++

poemonsense:
> > 1. This function seems not related to checkpoints. For all workloads, the list should work well. If so, then we should name it like `workload-list` instead of `ckpt-list`.
> > 2. Probably we need a CI for this test? The list could be, like, multiple microbench.bin lines.
> > 
> > Let @klin02 have a review and help the CI things
> 
> It's theoretically possible to support different workload lists, but for now I'm only supporting our new checkpoint resolution

What are the differences between checkpoints and other workloads? They are simply different images.

src/test/vsrc/vcs/top.v
```diff
@@ -197,8 +224,9 @@ reg [63:0] n_cycles;
 always @(posedge clock) begin
   if (reset) begin
     n_cycles <= 64'h0;
+    xs_rst_en<= 1'b0;
```
poemonsense, line 227: Avoid namings like this (`xs`). Difftest is not for XS only.

xiaokamikami:
> > > 1. This function seems not related to checkpoints. For all workloads, the list should work well. If so, then we should name it like `workload-list` instead of `ckpt-list`.
> > > 2. Probably we need a CI for this test? The list could be, like, multiple microbench.bin lines.
> > > 
> > > Let @klin02 have a review and help the CI things
> > 
> > 
> > It's theoretically possible to support different workload lists, but for now I'm only supporting our new checkpoint resolution
> 
> What are the differences between checkpoints and other workloads? They are simply different images.
Because the current limit for the next workload is based on the name in the checkpoint, if I need to run a regular list of applications, I can add a separate macro definition to distinguish between the two modes

poemonsense:
> > > > 1. This function seems not related to checkpoints. For all workloads, the list should work well. If so, then we should name it like `workload-list` instead of `ckpt-list`.
> > > > 2. Probably we need a CI for this test? The list could be, like, multiple microbench.bin lines.
> > > > 
> > > > Let @klin02 have a review and help the CI things
> > > 
> > > 
> > > It's theoretically possible to support different workload lists, but for now I'm only supporting our new checkpoint resolution
> > 
> > 
> > What are the differences between checkpoints and other workloads? They are simply different images.
> > Because the current limit for the next workload is based on the name in the checkpoint, if I need to run a regular list of applications, I can add a separate macro definition to distinguish between the two modes

I see. Currently this PR is using `sscanf(str1,"_%d_0.%ld_.gz", &checkpoin_idx, &max_instrs)` to extract the max-instrs.

I think we should avoid this. Probably we can use like `workload_path,max_instrs` in the workload list to indicate the max-instrs.

.github/workflows/main.yml
```diff
@@ -334,3 +334,14 @@ jobs:
             make simv DIFFTEST_PERFCNT=1 VCS=verilator -j2
             ./build/simv +workload=./ready-to-run/microbench.bin +e=0 +no-diff +max-cycles=100000
             ./build/simv +workload=./ready-to-run/microbench.bin +e=0 +diff=./ready-to-run/riscv64-nemu-interpreter-so
+
+      - name: Verilator Build with VCS Top (with workload list)
+        run : |
+            echo "TEST_HOME=/nfs/home/share/ci-workloads/workload-list-test" >> $GITHUB_ENV
```
poemonsense, line 340: Do not use internal server files in CI. CI runs on GitHub servers. We need to setup a workload list in CI and then use it.

For example:
```bash
echo "./ready-to-run/microbench.bin,10000" > list.txt
echo "./ready-to-run/linux.bin,20000" >> list.txt
./build/simv +workload-list=list.txt +diff=./ready-to-run/riscv64-nemu-interpreter-so
```

.github/workflows/main.yml
```diff
@@ -334,3 +334,14 @@ jobs:
...
+            cd $GITHUB_WORKSPACE/../xs-env
+            source ./env.sh
+            cd $GITHUB_WORKSPACE/../xs-env/NutShell
+            source ./env.sh
+            make clean
+            make simv DIFFTEST_PERFCNT=1 VCS=verilator -j2
+            ./build/simv  +workload=./ready-to-run/microbench.bin +e=0 +diff=./ready-to-run/riscv64-nemu-interpreter-so +max-instrs=10000 +workload-list=$TEST_HOME/list.txt
```
poemonsense, line 347: I think when `+workload-list` is given, we don't need `+workload` any more. Similarly, with `workload-list` specifying the max-instrs, we do not need to set `+max-instrs` again. Is it?

poemonsense: Also please add `\n` to the end of file. In the `Files changed` tab, GitHub shows a red mark here which means the code is missing the `\n` at the end of the file.

In VSCode, it is added by adding a new blank line at the end of the file. In vim, it should work well without explicit `\n`.

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -29,12 +29,17 @@
...
 
+#define STATE_LIMIT_EXCEEDED 3
+
 static bool has_reset = false;
 static char bin_file[256] = "ram.bin";
 static char *flash_bin_file = NULL;
 static bool enable_difftest = true;
 static uint64_t max_instrs = 0;
 
+static int ram_path_reset(char *file_name);
```
klin02, line 40: Seems no need to declare a func as `static` ?

klin02:
I have some questions about Memory reload and Software reload.
For memory:
1. Currently, we use $fread to load bin into synthesizable memory in initial block. When reset is set high again, this will not be acted again to reload DUT memory. See details in https://github.com/OpenXiangShan/difftest/blob/b4d9e66fe345007ca401ea9edeb1f1c98db81524/src/main/scala/common/Mem.scala#L41C14-L41C26.
2. Now we use synthesizable memory instead of DPIC by default. It's because DPIC with input and output will greatly slow down simulation at every calls. If you want to reload memory, maybe you can try extern DPIC function. Please think it carefully for clear structure.

For software:
It seems each simulation circuits can be seen like: begins with reset set to high (you will trigger simv_init in cycle0),ends with a non-zero simv_nstep result. 
1. Because simv_init will trigger several times, you should make sure previous resources is freed succesfully. Every circuits should be seperated with each other.
2. I think logic in top.v is mixed. Only a singal (seems like soft_rst_en) should be added to control reset. When a ckpt is ended, this signal will rise `reset` again. It will not affect any other logic.

Overall, I think the code structure will be like this:
1. When initial, you pass workload-list, and parse it to get workload array and max-instr array.
2. When reset is set to high. You will trigger simv_init, it will get `workload` and `max-instr`, just like only a workload is run. You will init software in this. And you may use extern DPIC function to load bin into DUT synthesizable memory.
3. When simv_nstep is ended successfully, and there is still remain ckpt. It will use a signal to set reset to high. And Software should be free clearly. Then Back to Point 2.
4. When last simv_nstep is ended (maybe return val is different from Point 3). The whole simulation is ended.

poemonsense:
It seems CI `test-difftest-vcs` fails but does not abort:

```
diff-test ref so:./ready-to-run/riscv64-nemu-interpreter-so
set checkpoint list path list.txt 
Can't open fp file 'list.txt'run work list start faild
- /home/runner/work/difftest/xs-env/NutShell/difftest/src/test/vsrc/vcs/top.v:138: Verilog $finish
```

Did you try the CI tests on local machine? You can test it with local servers using Verilator.

xiaokamikami:
> Better to rethink the structure carefully, It seems the logic can be simplified, and the feature should be compatible will existing features. Also, make sure the PR works well in Palladium. Because some feature (like memory load) is different with CI.

I have confirmed that it works well on Paladin (start DIPC). At present, I can run programs in batches with dpic faster than with rtl memory, so I do not add soft reset of Verilog memory for the time being

xiaokamikami:
> > It seems CI `test-difftest-vcs` fails but does not abort:
> > ```
> > diff-test ref so:./ready-to-run/riscv64-nemu-interpreter-so
> > set checkpoint list path list.txt 
> > Can't open fp file 'list.txt'run work list start faild
> > - /home/runner/work/difftest/xs-env/NutShell/difftest/src/test/vsrc/vcs/top.v:138: Verilog $finish
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Did you try the CI tests on local machine? You can test it with local servers using Verilator.
> 
> CI still seems to not work. Can you see the CI log? It seems the CI still does not run the workload list correctly. You can see the Checks tab to see the detailed logs.

Maybe I not create the file. Let me change it

klin02:
> > Better to rethink the structure carefully, It seems the logic can be simplified, and the feature should be compatible will existing features. Also, make sure the PR works well in Palladium. Because some feature (like memory load) is different with CI.
> 
> I have confirmed that it works well on Paladin (start DIPC). At present, I can run programs in batches with dpic faster than with rtl memory, so I do not add soft reset of Verilog memory for the time being

I would like to add some comments on DPIC memory. When running short workload (checkpoint), memory accessing times is so small that the cost maybe less than reload the whole memory to Hardware. I will add some option to `palladium.mk` for it.

xiaokamikami:
> > > Better to rethink the structure carefully, It seems the logic can be simplified, and the feature should be compatible will existing features. Also, make sure the PR works well in Palladium. Because some feature (like memory load) is different with CI.
> > 
> > 
> > I have confirmed that it works well on Paladin (start DIPC). At present, I can run programs in batches with dpic faster than with rtl memory, so I do not add soft reset of Verilog memory for the time being
> 
> I would like to add some comments on DPIC memory. When running short workload (checkpoint), memory accessing times is so small that the cost maybe less than reload the whole memory to Hardware. I will add some option to `palladium.mk` for it.

OK

> > > Better to rethink the structure carefully, It seems the logic can be simplified, and the feature should be compatible will existing features. Also, make sure the PR works well in Palladium. Because some feature (like memory load) is different with CI.
> > 
> > 
> > I have confirmed that it works well on Paladin (start DIPC). At present, I can run programs in batches with dpic faster than with rtl memory, so I do not add soft reset of Verilog memory for the time being
> 
> I would like to add some comments on DPIC memory. When running short workload (checkpoint), memory accessing times is so small that the cost maybe less than reload the whole memory to Hardware. I will add some option to `palladium.mk` for it.

Should I modify the macro definition according to your submission and then choose whether to enable soft reset function

klin02:
> Should I modify the macro definition according to your submission and then choose whether to enable soft reset function

Maybe you can directly modify palladium.mk
```Makefile
PLDM_MACRO_FLAGS 	+= +define+DIFFTEST +define+DISABLE_SIMJTAG_DPIC
ifneq ($(DIFFTEST_RAM_DPIC), 1)
PLDM_MACRO_FLAGS 	+= +define+DISABLE_DIFFTEST_RAM_DPIC
endif 
```
in https://github.com/OpenXiangShan/difftest/blob/b4d9e66fe345007ca401ea9edeb1f1c98db81524/palladium.mk#L17

Then when you compiler code in Palladium, you can use like `make pldm-build DIFFTEST_RAM_DPIC=1`

xiaokamikami:
> > Should I modify the macro definition according to your submission and then choose whether to enable soft reset function
> 
> Maybe you can directly modify palladium.mk
> 
> ```makefile
> PLDM_MACRO_FLAGS 	+= +define+DIFFTEST +define+DISABLE_SIMJTAG_DPIC
> ifneq ($(DIFFTEST_RAM_DPIC), 1)
> PLDM_MACRO_FLAGS 	+= +define+DISABLE_DIFFTEST_RAM_DPIC
> endif 
> ```
> 
> in
> 
> https://github.com/OpenXiangShan/difftest/blob/b4d9e66fe345007ca401ea9edeb1f1c98db81524/palladium.mk#L17
> 
> Then when you compiler code in Palladium, you can use like `make pldm-build DIFFTEST_RAM_DPIC=1`



> DISABLE_DIFFTEST_RAM_DPIC

OK

klin02 (approved):

.github/workflows/main.yml
```diff
@@ -334,3 +334,16 @@ jobs:
...
+            cd $GITHUB_WORKSPACE/../xs-env
+            source ./env.sh
+            cd $GITHUB_WORKSPACE/../xs-env/NutShell
+            source ./env.sh
+            make clean
+            echo "./ready-to-run/microbench.bin 10000" > list.txt
+            echo "./ready-to-run/linux.bin 20000" >> list.txt
+            make simv VCS=verilator WORKLOAD_SWITCH=1 -j2
+            ./build/simv +e=0 +diff=./ready-to-run/riscv64-nemu-interpreter-so +workload-list=./list.txt
+            rm -rf ./ready-to-run/list.txt
```
poemonsense, line 349: No need to remove.

klin02 (approved):

src/test/csrc/vcs/vcs_main.cpp
linkedlist771: Maybe u could try format tools like **black**.

poemonsense: We've adopted clang-format in the master branch. Thanks

src/test/vsrc/vcs/DeferredControl.v
```diff
@@ -31,7 +31,10 @@ initial $ixc_ctrl("sfifo", "set_deferred_result");
 `endif // PALLADIUM
 
 always @(posedge clock) begin
-  if (!reset && step != 0) begin
+  if (reset) begin
+    simv_result = 1'b0;
```
klin02, line 35: Use <= instead of =

src/test/vsrc/vcs/top.v
```diff
@@ -132,6 +132,11 @@ initial begin
   if ($test$plusargs("no-diff")) begin
     set_no_diff();
   end
+  // set exit instrs const
+  if ($test$plusargs("max-instrs")) begin
+    $value$plusargs("max-instrs=%d", max_instrs);
+    set_max_instrs(max_instrs);
```
klin02, line 138: I had changed this in another PR https://github.com/OpenXiangShan/difftest/pull/298

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -117,6 +117,7 @@ extern "C" uint8_t simv_init() {
     init_goldenmem();
     init_nemuproxy(DEFAULT_EMU_RAM_SIZE);
   }
+  simv_result = SIMV_RUN;
```
klin02, line 118: No need to change this? The simv_result is get from every simv_nstep. We declared it outside just becuase when some workload failed. No more simv_step should act.

poemonsense (approved):

PR #296, opened 2024-02-28
 CI: add clang-format with Github Actions

klin02:
We add clang-format to ensure consistent Cpp code style. When some code is no need to formatter, use clang-format off/on to control it.
Now our clang-format config is based on LLVM style, which can be checked by clang-format -style=LLVM --dump-config. 
All possible options and their meanings can be searched in https://clang.llvm.org/docs/ClangFormatStyleOptions.html


klin02:
clang-format will only resort consecutive `#include`, so we set `#include` controled by MACROS at the end for better code format.

poemonsense (approved):

klin02:
When user want to run `make clang-format` or `make format` locally. Please install clang-format first. 
Run `pip install --user clang-format`. And make sure `clang-format --version` is higher than 17.0.6 in your device.

PR #299, opened 2024-03-01
vcs: add ckpt result list dump to file

xiaokamikami:
When using the checkpoint list as the work load list as in the gem5 batch run, dump a list of performance statistics corresponding to the running program list according to the requirements of checkpoint score statistics.


klin02 (commented):
--workload-list option should be independent of the dump location. Now it's mixed together.

PR #300, opened 2024-03-01
vcs: fix resources deallocation of simulation

xiaokamikami:
When running with --workload-lilst, simv_init is reinitialized multiple times in a single startup, and the pointer from the last simv_init request needs to be destroyed before doing so. This commit fixed an issue with incomplete memory deallocation when exiting the program, and release the list file correctly when using the workload list

### DIFF
--- a/src/test/csrc/vcs/vcs_main.cpp
+++ b/src/test/csrc/vcs/vcs_main.cpp
@@ -82,9 +82,11 @@ int switch_workload() {
       set_max_instrs(num);
     } else if (feof(fp)) {
       printf("Workload list is completed\n");
+      fclose(fp);
       return 1;
     } else {
       printf("Unknown workload list format\n");
+      fclose(fp);
       return 1;
     }
   } else {
@@ -177,12 +179,21 @@ void difftest_deferred_result(uint8_t result) {
 }
 #endif // CONFIG_DIFFTEST_DEFERRED_RESULT
 
+void simv_finish() {
+  common_finish();
+  flash_finish();
+  if (enable_difftest) {
+    goldenmem_finish();
+    difftest_finish();
+  }
+  delete simMemory;
+  simMemory = nullptr;
+}
+
 static uint8_t simv_result = SIMV_RUN;
 #ifdef CONFIG_DIFFTEST_DEFERRED_RESULT
 extern "C" void simv_nstep(uint8_t step) {
-  if (simv_result == SIMV_FAIL)
-    return;
-  if (difftest == NULL)
+  if (simv_result == SIMV_FAIL || difftest == NULL)
     return;
 #else
 extern "C" uint8_t simv_nstep(uint8_t step) {
@@ -200,7 +211,7 @@ extern "C" uint8_t simv_nstep(uint8_t step) {
     int ret = simv_step();
     if (ret) {
       simv_result = ret;
-      difftest_finish();
+      simv_finish();
 #ifdef CONFIG_DIFFTEST_DEFERRED_RESULT
       difftest_deferred_result(ret);
       return;


klin02:
Wrap these finish functions into simv_finish, corresponding to simv_init.
By the way, fclose workload list can be also fixed in this PR？The title may be changed to vcs : free resources of simulation?

xiaokamikami:
> Wrap these finish functions into simv_finish, corresponding to simv_init. By the way, fclose workload list can be also fixed in this PR？The title may be changed to vcs : free resources of simulation?

Look at the static FILE* variable you used earlier. It's initialized only once, and you don't open a new file handle every time you call the function, so you don't need to change it

klin02:
> > Wrap these finish functions into simv_finish, corresponding to simv_init. By the way, fclose workload list can be also fixed in this PR？The title may be changed to vcs : free resources of simulation?
> 
> Look at the static FILE* variable you used earlier. It's initialized only once, and you don't open a new file handle every time you call the function, so you don't need to change it

For safety we still need to free the file handle at the end of the program. You don't need to free it in simv_finish（）, but in switch_workload（）

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -177,6 +177,13 @@ void difftest_deferred_result(uint8_t result) {
 }
 #endif // CONFIG_DIFFTEST_DEFERRED_RESULT
 
+void simv_finish() {
+  difftest_finish();
+  flash_finish();
```
klin02, line 182: simMemory should be also free, corresponding to `init_ram`, you can check the logic in emu.cpp.
Also, the logic should be the same as simv_init, as a general function. use enable_difftest as if condition.

xiaokamikami: ok

klin02:
https://github.com/OpenXiangShan/difftest/blob/bb991c9ee083c4e229c78cf1d4f2819f574c3852/src/test/csrc/verilator/emu.cpp#L503

src/test/csrc/common/ram.cpp
```diff
@@ -36,6 +36,12 @@ void init_ram(const char *image, uint64_t ram_size) {
   simMemory = new MmapMemory(image, ram_size);
 }
 
+void ram_finish() {
+#ifdef WITH_DRAMSIM3
```
klin02, line 40: It seems we will always call init_goldenmem in simv_init when enable_difftest?

src/test/csrc/common/ram.cpp
```diff
@@ -36,6 +36,12 @@ void init_ram(const char *image, uint64_t ram_size) {
...
+  dramsim3_finish();
+#endif
```
klin02, line 42: No need to declare another func? Like emu is OK

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -177,12 +179,20 @@ void difftest_deferred_result(uint8_t result) {
...
+  }
+}
+
 static uint8_t simv_result = SIMV_RUN;
 #ifdef CONFIG_DIFFTEST_DEFERRED_RESULT
 extern "C" void simv_nstep(uint8_t step) {
-  if (simv_result == SIMV_FAIL)
-    return;
-  if (difftest == NULL)
+  if (simv_result == SIMV_FAIL || difftest == NULL)
```
klin02, line 195: Is it necessary for changed?

xiaokamikami: It more concise

klin02: This change seems irrelevant to this PR. Such change may bring extra troubles for git. blame

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -177,12 +179,23 @@ void difftest_deferred_result(uint8_t result) {
...
 #endif // CONFIG_DIFFTEST_DEFERRED_RESULT
 
+void simv_finish() {
+  common_finish();
+  flash_finish();
+  if (enable_difftest) {
+    goldenmem_finish();
+    difftest_finish();
+  }
+#ifdef WITH_DRAMSIM3
```
klin02, line 189: dram_finish will call automatically with destructor of simMemory?
See https://github.com/OpenXiangShan/difftest/blob/bb991c9ee083c4e229c78cf1d4f2819f574c3852/src/test/csrc/common/ram.cpp#L298C1-L298C11

klin02:
> https://github.com/OpenXiangShan/difftest/blob/bb991c9ee083c4e229c78cf1d4f2819f574c3852/src/test/csrc/verilator/emu.cpp#L503

Click the link to see detail code. After delete, we should set it to nullptr.

klin02:
Try it in palladium with large size workload-list to test memory deallocation. Then I have no more problem.

xiaokamikami:
> Try it in palladium with large size workload-list to test memory deallocation. Then I have no more problem.

ok, I tested a set of workload-lists on palladium with runtime memory usage ranging from 6G to 16G without leaking memory

klin02 (approved):

PR #301, opened 2024-03-05
Support parseArgs for GatewayConfig

klin02:
We add an parseArgs interface of difftest for style and GatewayConfig. User can pass DifftestArgs from top module which extends App trait, and it will return filter args for sequent parsing.

Now, we support use DIFFTEST_STYLE=... to specify style such as DPIC. And use DIFFTEST_CONFIG=... to specify GatewayConfig, each letter represents an optimization measure.

### DIFF
--- a/src/main/scala/Difftest.scala
+++ b/src/main/scala/Difftest.scala
@@ -22,6 +22,7 @@ import difftest.gateway.{Gateway, GatewayConfig}
 
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Paths}
+import scala.annotation.tailrec
 import scala.collection.mutable.ListBuffer
 
 trait DifftestWithCoreid {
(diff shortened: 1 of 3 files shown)


klin02:
In Makefile of NutShell/XiangShan, a modify example like follows:
```
DIFFTEST_STYLE ?= dpic
DIFFTEST_CONFIG ?=
DIFFTEST_ARGS = DIFFTEST_STYLE=$(DIFFTEST_STYLE) DIFFTEST_CONFIG=$(DIFFTEST_CONFIG)
```
We need to append `$(DIFFTEST_ARGS)` to `mill` command.

src/main/scala/Difftest.scala
```diff
@@ -259,20 +259,38 @@ trait DifftestModule[T <: DifftestBundle] {
...
   private val vMacros = ListBuffer.empty[String]
 
+  def parseArgs(args: Array[String]): Array[String] = {
+    var newArgs = args
+    def parse(info: String): String = {
+      var target = ""
+      for (arg <- newArgs) {
+        if (arg.startsWith(info + "=")) {
+          newArgs = newArgs.filterNot(_ == arg)
+          target = arg.substring(info.length() + 1)
```
poemonsense, line 274: 1. Use `args.find(_.startsWith(info + "=")).map(_.substring(info.length() + 1)).getOrElse("dpic")`?

2. Line 279 sets the style. Then what's the default `var style = "dpic"` for? Will it be always overrided? What if the user does not specify any configs?

src/main/scala/Difftest.scala
```diff
@@ -282,21 +300,24 @@ object DifftestModule {
...
     val gateway = Gateway.collect()
     cppMacros ++= gateway.cppMacros
     vMacros ++= gateway.vMacros
     instances ++= gateway.instances
 
-    if (cppHeader.isDefined) {
-      generateCppHeader(cpu, cppHeader.get, gateway.structPacked.getOrElse(false))
+    val hasCppHeader = style match {
+      case "dpic" => true
+      case _      => false
```
poemonsense, line 317: Other styles can have C++ headers as well?

src/main/scala/Gateway.scala
```diff
@@ -94,8 +94,24 @@ object Gateway {
   private val instances = ListBuffer.empty[DifftestBundle]
   private var config = GatewayConfig()
 
-  def apply[T <: DifftestBundle](gen: T, style: String): T = {
-    config = GatewayConfig(style = style)
+  def setStyle(style: String): Unit = {
+    config = config.copy(style = style)
+  }
+
+  def setConfig(cfg: String): Unit = {
```
poemonsense, line 101: I think the current framework is not good enough. We are mixing the optimization passes and the communication methods. These configs optimize the difftest bundles. DPI-C (raw bundle, UInt, 4K bytes batch, ...) is about how the bundles/data are transferred.

But it's ok to merge this. We will come back in the future to re-think the design

poemonsense (dismissed):

poemonsense (approved):

PR #302, opened 2024-03-07
vcs: add dramsim on emu run

xiaokamikami:
Support for enabling dramsim3 programs to run on Paradine. DRAMSIM3_HOME Note: Recompile dramsim3 on Paladin and set DRAMSIM3_HOME, then add DRAMSIM3_HOME to LD_LIBRARY_PATH so that it links to the dynamic library correctly

### DIFF
--- a/src/test/csrc/common/ram.cpp
+++ b/src/test/csrc/common/ram.cpp
@@ -444,14 +444,19 @@ void dramsim3_init() {
 }
 
 void dramsim3_step() {
+  if (dram == NULL)
+    return;
   dram->tick();
 }
 
 void dramsim3_finish() {
   delete dram;
+  dram = NULL;
 }
 
 uint64_t memory_response(bool isWrite) {
+  if (dram == NULL)
+    return 0;
   auto response = (isWrite) ? dram->check_write_response() : dram->check_read_response();
   if (response) {
     auto meta = static_cast<dramsim3_meta *>(response->req->meta);
@@ -464,6 +469,8 @@ uint64_t memory_response(bool isWrite) {
 }
 
 bool memory_request(uint64_t address, uint32_t id, bool isWrite) {
+  if (dram == NULL)
+    return false;
   if (dram->will_accept(address, isWrite)) {
     auto req = new CoDRAMRequest();
     auto meta = new dramsim3_meta;
--- a/src/test/csrc/vcs/vcs_main.cpp
+++ b/src/test/csrc/vcs/vcs_main.cpp
@@ -179,6 +179,12 @@ void difftest_deferred_result(uint8_t result) {
 }
 #endif // CONFIG_DIFFTEST_DEFERRED_RESULT
 
+#ifdef WITH_DRAMSIM3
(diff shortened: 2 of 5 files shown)


src/test/csrc/common/ram.cpp
```diff
@@ -440,6 +440,8 @@ void dramsim3_init() {
 
   assert(dram == NULL);
   dram = new ComplexCoDRAMsim3(DRAMSIM3_CONFIG, DRAMSIM3_OUTDIR);
+  if (dram == NULL)
```
klin02, line 443: Is it necessary? Seems we call dramsim same as emu. No need to change the common interface.

xiaokamikami: Prevents references to null Pointers from causing Paladin to crash,If you don't think this is necessary, I can add a macro definition to enable checks on Paladin

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -199,6 +205,9 @@ extern "C" void simv_nstep(uint8_t step) {
 extern "C" uint8_t simv_nstep(uint8_t step) {
   if (difftest == NULL)
     return 0;
+#ifdef WITH_DRAMSIM3
```
klin02, line 208: If you want to trigger simv_tick every cycle. Sperate it for single interface, rather than leave it everywhere.

xiaokamikami: ok

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -179,6 +197,20 @@ void difftest_deferred_result(uint8_t result) {
 }
 #endif // CONFIG_DIFFTEST_DEFERRED_RESULT
 
+static uint8_t simv_result = SIMV_RUN;
+#ifdef WITH_DRAMSIM3
+extern "C" void simv_tick() {
+#ifdef CONFIG_DIFFTEST_DEFERRED_RESULT
```
klin02, line 203: No need to check simv_result and difftest here. You can check whether dram is NULL in dramsim3_step

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -190,7 +222,6 @@ void simv_finish() {
   simMemory = nullptr;
 }
 
-static uint8_t simv_result = SIMV_RUN;
```
klin02, line 193: No need to change here.

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -152,6 +162,14 @@ extern "C" uint8_t simv_step() {
   if (max_instrs != 0) { // 0 for no limit
     auto trap = difftest[0]->get_trap_event();
     if (max_instrs < trap->instrCnt) {
+      if (ckpt_item != NULL) {
```
klin02, line 165: Seems unexpected changes for dramsim? May you only push master branch on your fork repo

klin02 (approved):

PR #303, opened 2024-03-07
flash: support synthesizable flash without SYNTHESIS

klin02:
Previously we use synthesizable flash only when SYNTHESIS=1, now we support configuring synthesizable flash just like memory, so we can use it to run workload based on flash.


klin02:
By the way, I think we don't have to pass workload=... when flash is not NULL? If so, maybe I will open another PR to fix it.

poemonsense (approved):

PR #315, opened 2024-03-14
Add built-in Perf for difftest

klin02:
We will dump built-in Perf for difftest at the end of simulation. These printfs can be controled by difftest_perfCtrl_dump.

For emu, we call `Emulator::trigger_stat_dump`. For simv, we raise difftest_perfCtrl_dump when simv_result is not zero.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -304,7 +304,7 @@ jobs:
             cd $GITHUB_WORKSPACE/../xs-env/NutShell
             source ./env.sh
             make clean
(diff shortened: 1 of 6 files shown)


klin02:
To use built-in Perf in Palladium, we should delete `--disable-all` from `PLDM_ARGS` in XiangShan/Makefile. This option means remove all verilog `printf` in files, but it's allowed by Palladium and necessary for Perf. 
We may modify here when bump difftest, and test if there is any unexpected extra printfs in Palladium.

poemonsense (dismissed):

poemonsense (approved):

PR #320, opened 2024-03-19
Squash: tick squashed state independently between Groups

klin02:
We seperate DiffTestBundles into REF/GoldenMem, tick of Bundle will only affected by it's Group. Default Group is REF. Note some bundle will have several Group, because it will affect both REF and GoldenMem state. Bundles allowed to be dropped will ticked incidentially when it's Group tick.

We also add built-in perf to count the tick cause of each Group.


klin02:
TODO: For Replay, we need to copy DiffTraceInfo for different Squash Group, because different Group may be overlapped in squash range. We will implement replay by Group later: notify module `Replay` at certain range, and filter only replayed Group in `Squash`.

klin02:
Maybe we can use `pending` and `bump` to replace `want_tick` and `should_tick`

poemonsense (approved):

PR #322, opened 2024-03-22
ram: add gcpt-restore option for VCS

xiaokamikami:
The mobile gcpt overlay function becomes public so that it can also be used on the vcs Paladin

### DIFF
--- a/src/test/csrc/verilator/emu.cpp
+++ b/src/test/csrc/verilator/emu.cpp
@@ -382,10 +382,7 @@ Emulator::Emulator(int argc, const char *argv[])
   }
 
   if (args.gcpt_restore) {
-    char gcpt_restore[256];
-    uint64_t overwrite_nbytes = args.overwrite_nbytes;
-    strcpy(gcpt_restore, args.gcpt_restore);
-    overwrite_ram(gcpt_restore, overwrite_nbytes);
+    overwrite_ram(args.gcpt_restore, args.overwrite_nbytes);
   }
 
 #ifdef ENABLE_CHISEL_DB
--- a/src/test/csrc/verilator/emu.cpp
+++ b/src/test/csrc/verilator/emu.cpp
@@ -382,10 +382,7 @@ Emulator::Emulator(int argc, const char *argv[])
   }
 
   if (args.gcpt_restore) {
-    InputReader *reader = new FileReader(args.gcpt_restore);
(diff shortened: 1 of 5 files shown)


src/test/vsrc/vcs/top.v
```diff
@@ -126,6 +128,11 @@ initial begin
     $value$plusargs("flash=%s", flash_bin_file);
     set_flash_bin(flash_bin_file);
   end
+  // overwrite gcpt on ram: bin file
+  if ($test$plusargs("gcpt")) begin
```
poemonsense, line 132: Should use the same argument name as Verilator. Use gcpt-restore instead of gcpt.

poemonsense (approved):
Format the code. Then it looks ok to me

src/test/vsrc/vcs/top.v
xiaokamikami: ok

src/test/csrc/verilator/emu.cpp
```diff
@@ -382,10 +382,10 @@ Emulator::Emulator(int argc, const char *argv[])
   }
 
   if (args.gcpt_restore) {
-    InputReader *reader = new FileReader(args.gcpt_restore);
-    int overwrite_size = reader->read_all(simMemory->as_ptr(), args.overwrite_nbytes);
-    Info("Overwrite %d bytes from file %s.\n", overwrite_size, args.gcpt_restore);
-    delete reader;
+    char gcpt_restore[256];
+    uint64_t overwrite_nbytes = args.overwrite_nbytes;
+    strcpy(gcpt_restore, args.gcpt_restore);
```
poemonsense, line 387: why do we need this `strcpy`?

xiaokamikami: /home/runner/work/difftest/xs-env/NutShell/difftest/src/test/csrc/common/ram.h:217:6: note: candidate function not viable: 1st argument ('const char *') would lose const qualifier
There seems to be no way to pass the args argument as a function argument

poemonsense: should add const to the function declaration:

`void overwrite_ram(const char *gcpt_restore, uint64_t overwrite_nbytes)`

xiaokamikami: thankyou

poemonsense (approved):

PR #323, opened 2024-03-25
fix: gcpt over

xiaokamikami:
The override gcpt_restore operation should be performed before difftest is initialized, otherwise the override cannot be synchronized to the dut

### DIFF
--- a/src/test/csrc/vcs/vcs_main.cpp
+++ b/src/test/csrc/vcs/vcs_main.cpp
@@ -116,10 +116,11 @@ extern "C" uint8_t simv_init() {
   common_init("simv");
 
   init_ram(bin_file, DEFAULT_EMU_RAM_SIZE);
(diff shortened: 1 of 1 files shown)


src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -117,16 +117,15 @@ extern "C" uint8_t simv_init() {
 
   init_ram(bin_file, DEFAULT_EMU_RAM_SIZE);
   init_flash(flash_bin_file);
-
+  if (gcpt_restore_bin != NULL) {
```
poemonsense, line 120: put gcpt_restore after `init_ram` instead of `init_flash`?

xiaokamikami: ok

PR #325, opened 2024-03-28
Replay: seperate from Squash

klin02:
We seperate Replay feature for better structure. Signal in_replay will mark trace as replay or not, replayed trace will not be squashed. And we will record squashed range between trace_head and trace_tail, see squash function of DiffTraceInfo. In the future, Replay may support not only Squashed, but also Compress, etc.

This change also provides better flexibility for optimization. It will generated Cpp declaration depends on result of Gateway.

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -138,25 +138,30 @@ void difftest_squash_enable(int enable) {
   svSetScope(squashScope);
   set_squash_enable(enable);
 }
-
-#ifdef CONFIG_DIFFTEST_SQUASH_REPLAY
-extern "C" void set_squash_replay(int idx);
-void difftest_squash_replay(int idx) {
-  if (squashScope == NULL) {
-    printf("Error: Could not retrieve squash scope, set first\n");
-    assert(squashScope);
-  }
-  svSetScope(squashScope);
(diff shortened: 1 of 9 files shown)


src/test/csrc/difftest/difftest.cpp
```diff
@@ -165,32 +170,44 @@ Difftest::~Difftest() {
...
-    nFused_all += dut->commit[i].nFused;
+#ifdef CONFIG_DIFFTEST_REPLAY
+bool Difftest::can_replay() {
+  auto info = dut->trace_info;
+  printf("check %d %d %d\n", info.in_replay, info.trace_head, info.trace_tail);
+  return !info.in_replay && (info.trace_head != info.trace_tail);
+}
+
+bool Difftest::in_replay_range() {
+  printf("filter head %d tail %d\n", dut->trace_info.trace_head, dut->trace_info.trace_tail);
```
poemonsense, line 197: Are these printf's required?

poemonsense (approved):
Remove printf if it is not required

src/test/csrc/difftest/difftest.cpp
klin02: My mistake. I will delete it.

poemonsense (approved):

PR #328, opened 2024-04-02
ram: add dpic ram read_addr check

xiaokamikami:
I think it would be more efficient to also check for address bounds when reading memory to alert the user that the memory size might not match the RTL, instead of returning data at address 0 by using %
我认为应当在读内存时也对地址越界进行检查从而提醒用户内存大小可能设置与RTL不匹配，而不是靠取余返回0地址的数据，这会提高debug效率

This CHECK is enabled when PMEM CHECK is enabled


poemonsense (commented):
Originally we have this warning. But, later we find Linux kernel itself may read out-of-bound data. To avoid too many warnings, we remove the warning here.

If the user requires stricter RAM checking, we can enable this. How about adding a macro for this checker and printf? When the user wants this, they should enable it manually.

xiaokamikami:
> Originally we have this warning. But, later we find Linux kernel itself may read out-of-bound data. To avoid too many warnings, we remove the warning here.
> 
> If the user requires stricter RAM checking, we can enable this. How about adding a macro for this checker and printf? When the user wants this, they should enable it manually.

ok

poemonsense (approved):

PR #329, opened 2024-04-02
mem: add dramsim size config

xiaokamikami:
When dramsim is enabled, the memory size should be set to the same 16GB as for rlt
https://github.com/OpenXiangShan/XiangShan/blob/4a8a734e58e2469a23b631ed5ab41474c59ec5a1/src/test/scala/top/SimTop.scala#L45

### DIFF
--- a/config/config.h
+++ b/config/config.h
@@ -36,7 +36,11 @@
 // -----------------------------------------------------------------------
 
 // emulated memory size (Byte)
+#ifdef WITH_DRAMSIM3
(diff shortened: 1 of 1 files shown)


poemonsense:
I don't think this is a real issue that we need to fix. Memsize in scala is used to truncate the address. However, during simulation, difftest will decide which ram size to use. A smaller ram size will not affect the simulation. If the user requires a larger ram, the user should update the ram size at runtime using `--ram-size`

PR #330, opened 2024-04-03
Squash: use Queue to buffer unsquashed Bundle

klin02:
Now we use Queue to buffer StoreEvent, avoid too many squash tick
caused by it. This bundle will only compare store actions betweens
DUT and REF, no other bundles depends on it, which enable us to buffer
it.

Another bundle causing many ticks is Sbuffer, for it will update
GoldenMem, and other bundle related will affected by it. Now we still
cannot buffer and submitted such bundle together.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -362,6 +362,17 @@ jobs:
             ./build/simv +workload=./ready-to-run/microbench.bin +e=0 +no-diff +max-cycles=100000
             ./build/simv +workload=./ready-to-run/microbench.bin +e=0 +diff=./ready-to-run/riscv64-nemu-interpreter-so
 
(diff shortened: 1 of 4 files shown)


klin02:
This change requires NEMU expend STORE_QUEUE_SIZE to 128.
It speed in Palladium will increase from 300K to 484k

klin02:
Previous we will only submit latest Refill/Tlb/Load betweens Sbuffer and drop outdated. When enable flush feature, we will only submit Refill after latest Sbuffer in the Queue. More outdated Bundle relying on GoldenMem will be dropped.

Speed in Palladium will be increase to 534K.

But there are still some problems, dut to increase of BundleNum, the gates for Batch feature increase greatly, which cause verilator coredump, and consumes nearly 40M gates in Palladium.

poemonsense (approved):

PR #331, opened 2024-04-05
Batch: pack DPIC with hybird parallel and pipeline

klin02:
Previous we give each bundle a different delay to collect valid value by Pipeline.

This change process bundles of same class in parallel, because offset is easy to count, and it will reduce stages in pipeline to reduces regs.

We also sort bundles by width, giving larger bundle less delay for less reg. And add `num` to info, which will reduce info costs with multiple bundle of same class.


poemonsense (approved):
It would be better to add necessary code comments next time. Otherwise, we will be unable to understand the code in the future.

klin02:
Sure, I will leave more comments later, and a detailed doc when the code structure is more stable

PR #336, opened 2024-04-09
dramsim: hasdut memory_response

xiaokamikami:
On Paladin, the forward and backward DPIC may be in an indefinite order, preventing the queue from entering requests in an incorrect order by kicking out requests that are not in the correct order and returning false
need https://github.com/OpenXiangShan/XiangShan/pull/2859


poemonsense (changes requested):
Wait for further discussions before merge

klin02:
If we will have different func params for Palladium and other Platform, sepearte it. Same for difftest and RTL.

And param name will start with lowercase.

PR #339, opened 2024-04-11
Perf: report Simulation speed when finish

klin02:
As test in Palladium, Speed calculate by Time and CycleCnt is very close to Palladium report. We add this platform-independent report here.

### DIFF
--- a/src/test/csrc/common/perf.cpp
+++ b/src/test/csrc/common/perf.cpp
@@ -31,12 +31,14 @@ void difftest_perfcnt_init() {
   diffstate_perfcnt_init();
 }
 
-void difftest_perfcnt_finish() {
+void difftest_perfcnt_finish(uint64_t cycleCnt) {
   printf("==================== Difftest PerfCnt ====================\n");
   struct timespec ts;
   clock_gettime(CLOCK_MONOTONIC, &ts);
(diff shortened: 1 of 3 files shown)


src/test/csrc/common/perf.cpp
```diff
@@ -31,12 +31,14 @@ void difftest_perfcnt_init() {
...
 
-void difftest_perfcnt_finish() {
+void difftest_perfcnt_finish(uint64_t cycleCnt) {
   printf("==================== Difftest PerfCnt ====================\n");
   struct timespec ts;
   clock_gettime(CLOCK_MONOTONIC, &ts);
   perf_run_msec = ts.tv_sec * 1000 + ts.tv_nsec / 1000000 - perf_run_msec;
   printf("Run time: %lld s %lld ms\n", perf_run_msec / 1000, perf_run_msec % 1000);
+  float speed = (float)cycleCnt / (float)perf_run_msec;
+  printf("Speed: %.2f KHz\n", speed);
```
poemonsense, line 41: what about simulation speed? To avoid misunderstanding between simulation and cpu speed.

poemonsense (commented):
It would be better to use simulation speed instead of speed

poemonsense (approved):

PR #341, opened 2024-04-12
vcs: call difftest_finish also when no-diff

klin02:
This change is corresponding to difftest_init.


klin02:
Here `no-diff` means build with DIFFTEST, but run with `+no-diff`

poemonsense (approved):

PR #352, opened 2024-04-24
compress: using magic number determine file type

xyyy1420:


poemonsense:
I'm concerned that if a bin file starts with these bytes, what will happened?

Maybe when `NO_GZ_COMPRESSION` is defined, let this function always return `false`?

PR #353, opened 2024-04-24
Make: add support for compiling difftest as a lib.so

cailuoshan:
This commit adds support for using difftest as a lib to simulate with XS.
For now, it works both on VCS and Palladium.
use "make difftest-so" to generate libdifftest.so and link it when compiling your design.

### DIFF
--- a/libso.mk
+++ b/libso.mk
@@ -1,0 +1,18 @@
+LIBDIFFTEST_SO    	 = $(BUILD_DIR)/libdifftest.so
+CC_OBJ_DIR	 	 = $(abspath $(BUILD_DIR)/cc_obj)
+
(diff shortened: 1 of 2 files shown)


libso.mk
```diff
@@ -0,0 +1,18 @@
+LIBDIFFTEST_SO    	 = $(BUILD_DIR)/libdifftest.so
```
poemonsense, line 1: There seems to be more whitespaces than needed in the Makefile. Maybe remove some of them to keep it as fewer as possible? We can still align them using fewer whitespaces.

poemonsense:
Also, should we add a CI test to build the so? We can add a test in test-difftest-vcs

poemonsense (approved):

PR #355, opened 2024-04-25
stack: large the size

Maxpicca-Li:
MultiCore test in XiangShan will failed with origin small stack size. So there larges the stack size.

### DIFF
--- a/config/config.h
+++ b/config/config.h
@@ -170,7 +170,7 @@ extern unsigned long EMU_FLASH_SIZE;
 // #define ENABLE_IPC
 
 // stack size for emu
-#define EMU_STACK_SIZE (16 * 1024 * 1024)
+#define EMU_STACK_SIZE (32 * 1024 * 1024)
 
 // whether to maintain goldenmem
 #if NUM_CORES>1


Tang-Haojin (approved):

poemonsense:
It seems Verilator also checks the stack size (at least for latest versions). Do you check whether there's any public method to get the stack size from verilator build files? Then, we can use the public API to do this instead of always enlarging it every time xiangshan goes to bigger. For example, we cannot set the stack size to a very very big number just to simulate a 64-core xiangshan, even if we are only simulating a single-core xiangshan.

Maxpicca-Li:
@poemonsense Thank you for your reminder. I checked it out. On the Verilator official website, the only parameter related to adjusting the stack size that I could find is [`--no-unlimited-stack`](https://veripool.org/guide/latest/exe_verilator.html), which is used to disable the behavior of Verilator trying to disable the stack size limit using the `ulimit -s unlimited` command; it doesn't allow for adjusting the stack size. However, when compiling emu, we can add a `--stack-size` parameter just like `--seed` and `--max-instr`, to dynamically set the stack size which is used by emu. But I need to confirm the feasibility of the latter method. If you have any other more effective methods, please let me know. Thank you again.

poemonsense:
Looks good.

I remember when I run verilator, it warns me the stack size is too smaller. I will have a double check.

If you have free time, it's good if you can add the runtime argument. Thanks for your work

Maxpicca-Li:
Yes, there indeed was a warning when the stack size is too small. 

The image below is a screenshot of emu crashing due to a small stack size while debugging with gdb. However, I currently do not know how this was implemented. I need to check this further.

image

Maxpicca-Li:
> @poemonsense Thank you for your reminder. I checked it out. On the Verilator official website, the only parameter related to adjusting the stack size that I could find is [`--no-unlimited-stack`](https://veripool.org/guide/latest/exe_verilator.html), which is used to disable the behavior of Verilator trying to disable the stack size limit using the `ulimit -s unlimited` command; it doesn't allow for adjusting the stack size. However, when compiling emu, we can add a `--stack-size` parameter just like `--seed` and `--max-instr`, to dynamically set the stack size which is used by emu. But I need to confirm the feasibility of the latter method. If you have any other more effective methods, please let me know. Thank you again.

The latter method is implemented at the [PR 355](https://github.com/OpenXiangShan/difftest/pull/355).

PR #356, opened 2024-04-25
ram: fix zstd load on readFromZstd

xiaokamikami:
This will fix cases that may cause errors in other function segments because the new array is too large

### DIFF
--- a/src/test/csrc/common/compress.cpp
+++ b/src/test/csrc/common/compress.cpp
@@ -200,7 +200,7 @@ long readFromZstd(void *ptr, const char *file_name, long buf_size, uint8_t load_
 
   lseek(fd, 0, SEEK_SET);
 
-  long *temp_page = (long *)mallo(chunk_size);
+  long *temp_page = (long *)malloc(chunk_size);
   compress_file_buffer = (uint8_t *)malloc(file_size);
   assert(compress_file_buffer);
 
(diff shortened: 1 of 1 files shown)


src/test/csrc/common/compress.cpp
```diff
@@ -200,12 +200,14 @@ long readFromZstd(void *ptr, const char *file_name, long buf_size, uint8_t load_
 
   lseek(fd, 0, SEEK_SET);
 
-  compress_file_buffer = new uint8_t[file_size];
+  long *temp_page = (long *)mallo(chunk_size);
```
klin02, line 203: Pen-slip?

xiaokamikami: really,my fault

src/test/csrc/common/compress.cpp
```diff
@@ -200,12 +200,14 @@ long readFromZstd(void *ptr, const char *file_name, long buf_size, uint8_t load_
 
   lseek(fd, 0, SEEK_SET);
 
-  compress_file_buffer = new uint8_t[file_size];
+  long *temp_page = (long *)malloc(chunk_size);
```
xyyy1420, line 203: `new long[chunk_size]` alloc `chunk_size * sizeof(long)` bytes, but `malloc(chunk_size)` just alloc `chunk_size` bytes, If I understand correctly, there should be something wrong here

poemonsense:
For reviewers, please carefully format the squashed commit title/messages when merging this PR. Thanks

poemonsense:
This PR is due to some exceptions during VCS simulation. However, the C++ seems to have the same functionality before and after this PR. I'm converting this to draft now. After we find why the simulation aborts on VCS, we will come back.

PR #357, opened 2024-04-25
Squash: add SquashDependency for HCSR/VecCSR

klin02:
In difftest, CSR is only valid and used when instrcommit or exception. CSR value maybe invalid in other time, so we will only squash HCSR/VecCSR when there is also commit/event in this cycle.


pxk27 (approved):

poemonsense (approved):
I don't know much on the dependency. Feel free to merge it if you think it's good

PR #358, opened 2024-04-26
opt: dynamic set stack size of emu

Maxpicca-Li:
Solve [PR 355](https://github.com/OpenXiangShan/difftest/pull/355) more elegantly.

### DIFF
--- a/config/config.h
+++ b/config/config.h
@@ -170,7 +170,7 @@ extern unsigned long EMU_FLASH_SIZE;
 // #define ENABLE_IPC
 
 // stack size for emu
-#define EMU_STACK_SIZE (32 * 1024 * 1024)
+#define EMU_STACK_SIZE (16 * 1024 * 1024)
 
 // whether to maintain goldenmem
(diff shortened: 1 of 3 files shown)


poemonsense:
Thanks for the work. This PR looks good. However, I double check the progress of Verilator and find that it would be better to use the predicted stack size provided by Verilator. I'm proposing my ideas to Verilator in https://github.com/verilator/verilator/issues/4799.

If they accept to add the predictedStackSize method, we don't need this argument anymore. Otherwise, we can have this argument added.

Maxpicca-Li:
OK, that will be a better way to solve this.

poemonsense:
Our patch for Verilator has been accepted. Next release version of Verilator (v5.026) will come with a dynamic stack resizer using `setrlimit`. We've updated difftest to avoid explicit `setrlimit` in our own `main()` function in https://github.com/OpenXiangShan/difftest/pull/367. This PR is closed. Thanks all for this work.

PR #359, opened 2024-04-26
vcs: check max-instrs for every cpu core

xiaokamikami:
When exiting the program using max_instrs, ensure that each core runs to the corresponding number of instructions, and add the core id when printing the IPC

### DIFF
--- a/src/test/csrc/vcs/vcs_main.cpp
+++ b/src/test/csrc/vcs/vcs_main.cpp
@@ -43,11 +43,11 @@ static bool enable_difftest = true;
 static uint64_t max_instrs = 0;
 static char *workload_list = NULL;
 static uint64_t overwrite_nbytes = 0xe00;
-struct CORE_END_INFO {
+struct core_end_info_t {
   bool core_trap[NUM_CORES];
   uint8_t core_trap_num;
 };
-static CORE_END_INFO core_end_info;
+static core_end_info_t core_end_info;
 
 enum {
   SIMV_RUN,
@@ -197,10 +197,10 @@ extern "C" uint8_t simv_step() {
   if (max_instrs != 0) { // 0 for no limit
     for (int i = 0; i < NUM_CORES; i++) {
       if (core_end_info.core_trap[i])
-        break;
+        continue;
       auto trap = difftest[i]->get_trap_event();
       if (max_instrs < trap->instrCnt) {
-        core_end_info.core_trap[i] = 1;
+        core_end_info.core_trap[i] = true;
         core_end_info.core_trap_num++;
         eprintf(ANSI_COLOR_GREEN "EXCEEDED CORE-%d MAX INSTR: %ld\n" ANSI_COLOR_RESET, i, max_instrs);
         difftest[i]->display_stats();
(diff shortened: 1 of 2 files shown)


src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -190,11 +195,18 @@ extern "C" uint8_t simv_step() {
...
 
   if (max_instrs != 0) { // 0 for no limit
-    auto trap = difftest[0]->get_trap_event();
-    if (max_instrs < trap->instrCnt) {
-      eprintf(ANSI_COLOR_GREEN "EXCEEDED MAX INSTR: %ld\n" ANSI_COLOR_RESET, max_instrs);
-      difftest[0]->display_stats();
-      return SIMV_DONE;
+    for (int i = 0; i < NUM_CORES; i++) {
+      if (core_end_info.core_trap[i])
+        break;
```
poemonsense, line 200: Why does it `break`? Shouldn't it be `continue`?

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -43,6 +43,11 @@ static bool enable_difftest = true;
 static uint64_t max_instrs = 0;
 static char *workload_list = NULL;
 static uint64_t overwrite_nbytes = 0xe00;
+struct CORE_END_INFO {
```
poemonsense, line 46: We can use another name for this struct, because C++ normally don't have a type with all capital letters. Maybe `core_end_info_t`?

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -190,11 +195,18 @@ extern "C" uint8_t simv_step() {
...
-    if (max_instrs < trap->instrCnt) {
-      eprintf(ANSI_COLOR_GREEN "EXCEEDED MAX INSTR: %ld\n" ANSI_COLOR_RESET, max_instrs);
-      difftest[0]->display_stats();
-      return SIMV_DONE;
+    for (int i = 0; i < NUM_CORES; i++) {
+      if (core_end_info.core_trap[i])
+        break;
+      auto trap = difftest[i]->get_trap_event();
+      if (max_instrs < trap->instrCnt) {
+        core_end_info.core_trap[i] = 1;
```
poemonsense, line 203: Why don't we use true and false?

poemonsense (commented):
The major difference here with the prior version is that, this requires all cores to exceed the max-instr count to finish the simulation. So, is there any reason that we have to fix this? I'm not familiar with multicore CPU behaviors.

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -190,11 +195,18 @@ extern "C" uint8_t simv_step() {
   }
 
   if (max_instrs != 0) { // 0 for no limit
-    auto trap = difftest[0]->get_trap_event();
-    if (max_instrs < trap->instrCnt) {
-      eprintf(ANSI_COLOR_GREEN "EXCEEDED MAX INSTR: %ld\n" ANSI_COLOR_RESET, max_instrs);
-      difftest[0]->display_stats();
-      return SIMV_DONE;
+    for (int i = 0; i < NUM_CORES; i++) {
```
poemonsense, line 198: I'm also wondering why do we need this core_end_info. If we check every core each time, it works as well.

bool is_finished = true;
for () { if (!exceed) is_finished = false; }

This code also works and looks simpler.

PR #362, opened 2024-04-28
Makefile: check clang-format version

klin02:
Add Version Check and Install note into Makefile. So user should ensure correct clang-format version for difftest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ SIM_CXXFLAGS += -DNOOP_HOME=\\\"$(NOOP_HOME)\\\"
 
 # generated-src
 GEN_CSRC_DIR  = $(BUILD_DIR)/generated-src
-SIM_CXXFILES += $(shell find $(GEN_CSRC_DIR) -name "*.cpp")
+SIM_CXXFILES += $(shell find $(GEN_CSRC_DIR) -name "*.cpp" 2> /dev/null)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -212,7 +212,14 @@ format: scala-format clang-format
 scala-format:
 	mill -i mill.scalalib.scalafmt.ScalafmtModule/reformatAll __.sources
 
+CLANG_FORMAT_VER=18_1_4
```
poemonsense, line 215: whitespaces before and after `=`

poemonsense (commented):
When clang-format does not exist, make anything will cause a `/bin/sh: 1: clang-format: not found`. `make clang-format` will cause `/bin/sh: 1: clang-format: not found` and `make: clang-format: No such file or directory`. They need to be avoided.

poemonsense:
To avoid sh warning, may add something like this:

```diff
-SIM_CXXFILES += $(shell find $(GEN_CSRC_DIR) -name "*.cpp")
+SIM_CXXFILES += $(shell find $(GEN_CSRC_DIR) -name "*.cpp" 2> /dev/null)
```

To avoid make warning, may check its existence before running --version

poemonsense (commented):
Why do we require a specific `clang-format` version?

poemonsense (approved):

PR #363, opened 2024-04-29
difftest: support dual warmup on all platforms

xiaokamikami:
Clearing all performance counters with difftest_perfCtrl_clean can lead to incorrect performance statistics on multiple cores, because the number of instruction commits varies from core to core
Now, the information of the completion of warmup under multi-core is counted into the difftest framework, which supports several platforms currently used

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -1171,7 +1171,7 @@ void Difftest::display_stats() {
   auto trap = get_trap_event();
   uint64_t instrCnt = trap->instrCnt;
   uint64_t cycleCnt = trap->cycleCnt;
-  double ipc = (double)instrCnt / cycleCnt;
+  double ipc = (double)(instrCnt - warmup_instrs) / (cycleCnt - warmup_cycle);
   eprintf(ANSI_COLOR_MAGENTA "Core-%d instrCnt = %'" PRIu64 ", cycleCnt = %'" PRIu64 ", IPC = %lf\n" ANSI_COLOR_RESET,
           this->id, instrCnt, cycleCnt, ipc);
 }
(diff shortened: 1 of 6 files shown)


src/test/csrc/difftest/difftest.cpp
```diff
@@ -1171,7 +1171,7 @@ void Difftest::display_stats() {
   auto trap = get_trap_event();
   uint64_t instrCnt = trap->instrCnt;
   uint64_t cycleCnt = trap->cycleCnt;
-  double ipc = (double)instrCnt / cycleCnt;
+  double ipc = (double)(instrCnt - warmup_instrs) / (cycleCnt - warmup_cycle);
```
poemonsense, line 1174: Warmup is for the performance counters only. This IPC is the report for overall simulation. We do not consider the warmup here.

src/test/csrc/difftest/difftest.h
```diff
@@ -273,6 +273,11 @@ class Difftest {
     state->dump_commit_trace = enable;
   }
 
+  void set_warmup_info(uint64_t cycle, uint64_t instr) {
```
poemonsense, line 276: warmup is not related to difftest. Thus, they should not be in difftest

poemonsense (commented):
It's ok to use the difftest cycleCnt for the IPC (which is one of the performance counters the dut may provide). It's worth noting only IPC can be got from difftest.

Still, warmup is for the performance counters from the DUT. If you require different control signals for different cores, we should change how the performance counters are handled. This is not only about the IPC, but the performance counters.

xiaokamikami:
> It's ok to use the difftest cycleCnt for the IPC (which is one of the performance counters the dut may provide). It's worth noting only IPC can be got from difftest.
> 
> Still, warmup is for the performance counters from the DUT. If you require different control signals for different cores, we should change how the performance counters are handled. This is not only about the IPC, but the performance counters.

So should I support ports for different cores to clear performance counters independently in hardware with multiple cores?

PR #365, opened 2024-05-06
Fix difftest build in non-XiangShan environment

forever043:

PR #366, opened 2024-05-08
CI: use the default NOOP_HOME for difftest_verilog

poemonsense:


poemonsense:
#365 fixes the export of default `NOOP_HOME`. CI no longer requires it.

PR #379, opened 2024-05-15
difftest: allow early store events for the current instruction

good-circle:


poemonsense (approved):

poemonsense:
This update requires the store event to be updated to difftest no earlier than the commitment of current store instruction.

poemonsense:
Github "squash and merge" lets the PR author be the commit author. To avoid such cases, please let the original commit author submit the PR

PR #381, opened 2024-05-20
assert: add v2 to include filename information

poemonsense:
This commit adds the xs_assert_v2 and introduces the `filename` argument to show which file the assertion line lies. This is very useful if the design is generated using `--split-verilog`.

For backward compatibility, xs_assert is preserved.

### DIFF
--- a/src/test/csrc/common/common.cpp
+++ b/src/test/csrc/common/common.cpp
@@ -22,9 +22,18 @@ int assert_count = 0;
 int signal_num = 0;
 const char *emu_path = NULL;
 
+// Usage in SV/Verilog: xs_assert(`__LINE__);
 extern "C" void xs_assert(long long line) {
   if (assert_count >= 0) {
     printf("Assertion failed at line %lld.\n", line);
+    assert_count++;
+  }
+}
+
+// Usage in SV/Verilog: xs_assert_v2(`__FILE__, `__LINE__);
+extern "C" void xs_assert_v2(const char *filename, long long line) {
+  if (assert_count >= 0) {
+    printf("Assertion failed at %s:%lld.\n", filename, line);
     assert_count++;
   }
 }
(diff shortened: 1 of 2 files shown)


poemonsense:
I'm going to update rocket-chip and nutshell to use `xs_assert_v2` instead of `xs_assert` to show filename when assertion fails.

@Tang-Haojin If XiangShan is going to use `--split-verilog` in the future, you may try using v2 as well.

```diff
-               sed -i -e 's/$$fatal/xs_assert(`__LINE__)/g' "$$file";           \
+               sed -i -e 's/$$fatal/xs_assert_v2(`__FILE__, `__LINE__)/g' "$$file";           \
```

poemonsense:
The expected output looks like:


```
➜  NutShell git:(master) ✗ ./build/emu -i ./ready-to-run/microbench.bin
emu compiled at May 20 2024, 10:40:21
Using simulated 32768B flash
Using simulated 8192MB RAM
The image is ./ready-to-run/microbench.bin
The reference model is /home/xuyinan/nutshell2/NEMU/build/riscv64-nemu-interpreter-so
The first instruction of core 0 has commited. Difftest enabled. 
nemu does not enable flash fetch!
Assertion failed at /home/xuyinan/nutshell2/NutShell/build/rtl/WBU.sv:137.
The simulation stopped. There might be some assertion failed.
Core 0: ABORT at pc = 0x8000464c
Core-0 instrCnt = 4, cycleCnt = 552, IPC = 0.007246
Seed=0 Guest cycle spent: 553 (this will be different from cycleCnt if emu loads a snapshot)
Host time spent: 2ms
Assertion failed
    at WBU.scala:60 assert(!difftest_commit.valid || difftest_commit.pc < "h80001000".U)
```

klin02 (approved):

poemonsense:
There is a `if (assert_count >= 0)` in the code. But the `assert_count` is never set to a negavtive value.

The bug seems to be introduced in https://github.com/OpenXiangShan/difftest/commit/41f83845484ffb43c3c3d9b5b0b7c30a7fc7b1f2. I'll fix it in another PR. This bug will make the resetGen simulation fail.

PR #382, opened 2024-05-20
common: fix the assertion control logic

poemonsense:
This commit fixes the assertion control logic by introducing a init function without setting assertion_count to 0. The user may manually call common_enable_assert to enable assertions.

Previously we are checking whether the assertions are allowed by negative assert_count values. However, this variable is incorrectly set to zero during common_init.

### DIFF
--- a/src/test/csrc/common/common.cpp
+++ b/src/test/csrc/common/common.cpp
@@ -62,7 +62,7 @@ uint32_t uptime(void) {
 
 static char mybuf[BUFSIZ];
 
-void common_init(const char *program_name) {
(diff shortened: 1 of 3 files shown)


poemonsense:
VCS still uses the common_init function which allows assertions by default.

For verilator, we are separating the common_init stage and assertion init stage. Assertions are allowed during the RTL reset period. The bug is introduced in https://github.com/OpenXiangShan/difftest/commit/41f83845484ffb43c3c3d9b5b0b7c30a7fc7b1f2

klin02 (approved):

PR #383, opened 2024-05-20
vcs: Move difftest control logic to separated DifftestEndpoint module

forever043:
The difftest control logic was independent of testbench, move it to a separated module `DifftestEndpoint`. The clock/reset/fsdb related logic, which were strongly related to testbench, were kept in the orignal tb_top module.

Also add BOSC to the copyright of modified files.


poemonsense:
I think it's good to separate difftest logic to a new module. This makes the TOP cleaner.

src/test/vsrc/vcs/DifftestEndpoint.v
```diff
@@ -0,0 +1,241 @@
...
+  // log end
+  if ($test$plusargs("e")) begin
+    $value$plusargs("e=%d", difftest_logCtrl_end_r);
+  end
+  else begin
+    difftest_logCtrl_end_r = 0;
+  end
+`ifndef TB_NO_DPIC
+  // workload: bin file
+  if ($test$plusargs("workload")) begin
```
poemonsense, line 93: Some arguments (such as +workload, +flash, ...) may not depend on the existence of difftest. We may move them into a separate module only for simulation arguments. Then, in the future, we can instantiate this DifftestEndpoint only if we need difftest (for crosscheck against a REF).

This will require a lot of engineering work. We can do it in future PRs.

poemonsense:
@klin02 please help review whether the changes are good and do not bring unexpected issues

poemonsense:
For WIP PR, we can convert it to a draft on the GitHub webpage under the "Reviewers" tab

klin02 (approved):

PR #387, opened 2024-05-30
difftest: move loadevent smp instruction

xiaokamikami:
The load instruction was moved to a separate function, and the debug print in the function could not be compiled was fixed, and the repeated judgment statements were reduced

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -505,12 +505,12 @@ int Difftest::do_instr_commit(int i) {
   }
 #if NUM_CORES > 1
   // Handle load instruction carefully for SMP
-  load_instruction(i, realWen);
+  do_load_smp_check(i, realWen);
 #endif
   return 0;
 }
 
-void Difftest::load_instruction(int i, bool realWen) {
+void Difftest::do_load_smp_check(int i, bool realWen) {
 #ifdef CONFIG_DIFFTEST_LOADEVENT
   if (dut->load[i].isLoad || dut->load[i].isAtomic) {
     proxy->sync();
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -503,85 +503,92 @@ int Difftest::do_instr_commit(int i) {
   for (int j = 0; j < dut->commit[i].nFused; j++) {
     proxy->ref_exec(1);
   }
-
+#if NUM_CORES > 1
   // Handle load instruction carefully for SMP
-  if (NUM_CORES > 1) {
-#ifdef CONFIG_DIFFTEST_LOADEVENT
-    if (dut->load[i].isLoad || dut->load[i].isAtomic) {
-      proxy->sync();
-      bool reg_cmp_fail =
-          !dut->commit[i].vecwen && *proxy->arch_reg(dut->commit[i].wdest, dut->commit[i].fpwen) != get_commit_data(i);
-      if (realWen && reg_cmp_fail) {
-        // printf("---[DIFF Core%d] This load instruction gets rectified!\n", this->id);
-        // printf("---    ltype: 0x%x paddr: 0x%lx wen: 0x%x wdst: 0x%x wdata: 0x%lx pc: 0x%lx\n", dut->load[i].opType, dut->load[i].paddr, dut->commit[i].wen, dut->commit[i].wdest, get_commit_data(i), dut->commit[i].pc);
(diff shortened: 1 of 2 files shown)


src/test/csrc/difftest/difftest.h
```diff
@@ -215,6 +215,7 @@ class Difftest {
   bool has_commit = false;
   // Trigger a difftest checking procdure
   int step();
+  void load_instruction(int i, bool realWen);
```
poemonsense, line 218: move it to a private member and rename it with `do_load_check` or similar names as other functions

src/test/csrc/difftest/difftest.cpp
```diff
@@ -503,85 +503,92 @@ int Difftest::do_instr_commit(int i) {
...
-          difftest[(NUM_CORES - 1) - this->id]->proxy->memcpy(dut->load[i].paddr, &buf, len, DIFFTEST_TO_DUT);
-          Info("---    content: %lx\n", buf);
-#else
-          proxy->ref_memcpy(dut->load[i].paddr, &golden, len, DUT_TO_REF);
-          if (realWen) {
-            if (!dut->commit[i].vecwen) {
-              *proxy->arch_reg(dut->commit[i].wdest, dut->commit[i].fpwen) = get_commit_data(i);
-            }
-            proxy->sync(true);
+      printf("---[DIFF Core%d] This load instruction gets rectified!\n", this->id);
```
poemonsense, line 552: It seems line 552-556 are newly added in this PR. Why do we need this? I don't see them in the original code.

src/test/csrc/difftest/difftest.cpp
```diff
@@ -503,85 +503,92 @@ int Difftest::do_instr_commit(int i) {
...
+      printf("---    ltype: 0x%x paddr: 0x%lx wen: 0x%x wdst: %d wdata: 0x%lx pc: 0x%lx\n", dut->load[i].opType,
+             dut->load[i].paddr, dut->commit[i].wpdest, dut->commit[i].wdest, commitData, dut->commit[i].pc);
+      printf("---    nFuse %d\n", dut->commit[i].nFused);
+      printf("---    golden: 0x%lx  ref: 0x%lx  arch_reg: 0x%lx\n", golden, refArchRegData, dutArchRegData);
+#endif
+      if (golden == commitData) {
+        proxy->ref_memcpy(dut->load[i].paddr, &golden, len, DUT_TO_REF);
+        if (dutArchRegData == refArchRegData && dutArchRegData != commitData) {
```
poemonsense, line 560: Why is there `dutArchRegData == refArchRegData && dutArchRegData != commitData`?

I check the original code and find there is `realWen`. Should it be `realWen`?

src/test/csrc/difftest/difftest.cpp
```diff
@@ -503,85 +503,92 @@ int Difftest::do_instr_commit(int i) {
...
+          Info("---    ref golden == commitdate bud not == dut reg, need to check difftest");
```
poemonsense, line 561: Why do we need this `Info`? I don't see this line in the original code.

src/test/csrc/difftest/difftest.cpp
```diff
@@ -503,85 +503,92 @@ int Difftest::do_instr_commit(int i) {
...
+#ifdef DEBUG_SMP
+        // goldenmem check failed as well, raise error
+        Info("---  SMP difftest mismatch!\n");
+        Info("---  Trying to probe local data of another core\n");
+        uint64_t buf;
+        difftest[(NUM_CORES - 1) - this->id]->proxy->ref_memcpy(dut->load[i].paddr, &buf, len, REF_TO_DUT);
+        Info("---    content: %lx\n", buf);
 #endif
+        proxy->ref_memcpy(dut->load[i].paddr, &golden, len, DUT_TO_REF);
+        if (!dut->commit[i].vecwen) {
```
poemonsense, line 584: There is a missing `if (realWen) {` compared to the original version. Why do we delete it?

src/test/csrc/difftest/difftest.cpp
poemonsense: The argument `realWen` seems unused in this new function.

poemonsense (commented):
I agree separating the load checker is good for us.

However, the current changes seem not the same as the original version. We need to clarify the differences.

klin02:
For multicore difftest, loadEvent will be seperated from instrCommit for more flexible logic. And some logic will be also not the same as currently. Maybe some format change should be also changed later

PR #388, opened 2024-05-30
squash: use soft queue to buffer unsquashed Diffstate

klin02:
Now there are some diffstates still can not squash, such as StoreEvent, which will affect squash of related event.

Previous we use hardware queue to buffer such event for strict order. For less gates consumed, larger queue size and easier order control, now we try to buffer such event by soft queue.

In hardware, we should continue squashing related event while submit unsquash ones. Software logic should also correspond to this, may be depends on some MACRO.

We also fix wrong process order of StoreEvent. Since all StoreEvent submitted before has been enqueued. Check should be after all InstrCommits.


klin02:
Now, StoreEvent is most important reason for squash break. And it seems more easy to handle, only related to instrCommits.
We will first try to deal with StoreEvent. And try more common logic for more events.

klin02:
We deal with unsquashed DiffState in PR #405. So close this PR

PR #389, opened 2024-05-31
difftest: force load unsquash for multicore, reset after use

klin02:
Load is related to reg writeback for goldenMem and regfile compare. Previous we ignore load for singleCore for better squash rate. But currently load is unsquash, because different core relys on load to pass reg data from other cores. We use load data to change REF model regfile.

And we also reset isLoad and isAtomic after use. These signals act as if-cond, and outdated ones will result in wrong reg copy.


klin02:
BTW，logic in Squasher module seems a little messing. We will try to wrapper logic by some Object, in order to better handle with Vec[DiffState] squashing.

poemonsense (approved):

PR #390, opened 2024-05-31
squash: fix force-sync of first-commit for multicore

klin02:
First commit of multicore is not valid in same cycle, previous single control logic only works for single core.

This change mark init state for each core, first commits of every core will be force-sync.

### DIFF
--- a/src/main/scala/Squash.scala
+++ b/src/main/scala/Squash.scala
@@ -128,12 +128,17 @@ class Squasher(bundleType: DifftestBundle, length: Int, numCores: Int, config: G
 
   // Mark the initial commit events as non-squashable for initial state synchronization.
   val tick_first_commit = Option.when(bundleType.desiredCppName == "commit") {
-    val hasValidCommitEvent = VecInit(state.map(_.bits.getValid).toSeq).asUInt.orR
-    val isInitialEvent = RegInit(true.B)
-    when(isInitialEvent && hasValidCommitEvent) {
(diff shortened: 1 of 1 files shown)


klin02:
This PR based on PR #389 . I will rebase after that PR merged into master and delete this comment.

src/main/scala/Squash.scala
```diff
@@ -128,12 +128,14 @@ class Squasher(bundleType: DifftestBundle, length: Int, numCores: Int, config: G
 
   // Mark the initial commit events as non-squashable for initial state synchronization.
   val tick_first_commit = Option.when(bundleType.desiredCppName == "commit") {
-    val hasValidCommitEvent = VecInit(state.map(_.bits.getValid).toSeq).asUInt.orR
-    val isInitialEvent = RegInit(true.B)
-    when(isInitialEvent && hasValidCommitEvent) {
-      isInitialEvent := false.B
-    }
-    isInitialEvent && hasValidCommitEvent
+    VecInit((0 to numCores).map { id =>
```
poemonsense, line 131: VecInit(Seq.tabulate(numCores)(id => ...)).asUInt.orR

poemonsense: https://www.chisel-lang.org/api/latest/chisel3/VecInit$.html

Maybe `VecInit.tabulate(numCores)(id => ...)`

poemonsense (approved):

PR #392, opened 2024-06-07
verilator.mk: Add support for PGO

cyyself:
This commit add PGO_CFLAGS and PGO_LDFLAGS when calling verilator to generate Makefile. And add script to build PGO for both GCC and LLVM.

The following is the usage of PGO:

```shell
make emu MFC=1 EMU_THREADS=16 NUM_CORES=1 PGO_WORKLOAD=`realpath ready-to-run/coremark-2-iteration.bin` LLVM_PROFDATA=llvm-profdata -j `nproc`
```

Note: The LLVM_PROFDATA is hard to detect if multiple versions of LLVM are installed. An example is that on Debian, we can install different versions of LLVM and use a version with suffixes like `clang++-18`. The LLVM_PROFDATA should be set to the correct version like `llvm-profdata-18` in this case.

If the compiler of verilator is GCC, we should not use LLVM_PROFDATA:

```shell
make emu MFC=1 EMU_THREADS=16 NUM_CORES=1 PGO_WORKLOAD=`realpath ready-to-run/coremark-2-iteration.bin` -j `nproc`
```

### DIFF
--- a/verilator.mk
+++ b/verilator.mk
@@ -105,9 +105,11 @@ VERILATOR_FLAGS =                   \
   --output-split 30000              \
   --output-split-cfuncs 30000       \
   -I$(RTL_DIR)                      \
-  -I$(GEN_VSRC_DIR)		    \
+  -I$(GEN_VSRC_DIR)                 \
   -CFLAGS "$(EMU_CXXFLAGS)"         \
   -LDFLAGS "$(EMU_LDFLAGS)"         \
+  -CFLAGS "\$$(PGO_CFLAGS)"         \
+  -LDFLAGS "\$$(PGO_LDFLAGS)"       \
   -o $(abspath $(EMU))              \
   $(VEXTRA_FLAGS)
 
@@ -117,6 +119,10 @@ EMU_DEPS  := $(SIM_VSRC) $(EMU_CXXFILES)
 EMU_HEADERS := $(shell find $(EMU_CSRC_DIR) -name "*.h")     \
                $(shell find $(SIM_CSRC_DIR) -name "*.h")     \
                $(shell find $(DIFFTEST_CSRC_DIR) -name "*.h")
+
+# Profile Guided Optimization
+EMU_PGO_DIR  = $(EMU_DIR)/pgo
+PGO_MAX_CYCLE ?= 2000000
 
 $(EMU_MK): $(SIM_TOP_V) | $(EMU_DEPS)
 ifeq ($(EMU_COVERAGE),1)
@@ -138,13 +144,48 @@ EMU_COMPILE_FILTER =
 build_emu_local: $(EMU_MK)
 	@echo -e "\n[c++] Compiling C++ files..." >> $(TIMELOG)
 	@date -R | tee -a $(TIMELOG)
+ifdef PGO_WORKLOAD
+	@echo "Building PGO profile..."
+	@stat $(PGO_WORKLOAD) > /dev/null
+	@$(MAKE) clean_obj
+	@mkdir -p $(EMU_PGO_DIR)
+	$(TIME_CMD) $(MAKE) -s VM_PARALLEL_BUILDS=1 OPT_FAST="-O3" OPT_SLOW="-O0" PGO_CFLAGS="-fprofile-generate=$(EMU_PGO_DIR)" PGO_LDFLAGS="-fprofile-generate=$(EMU_PGO_DIR)" -C $(<D) -f $(<F) $(EMU_COMPILE_FILTER)
+	@echo "Training emu with PGO Workload..."
+	$(EMU) -i $(PGO_WORKLOAD) --max-cycles=$(PGO_MAX_CYCLE) 1>$(EMU_PGO_DIR)/`date +%s`.log 2>$(EMU_PGO_DIR)/`date +%s`.err $(PGO_EMU_ARGS)
+ifdef LLVM_PROFDATA
+# When using LLVM's profile-guided optimization, the raw data can not
+# directly be used in -fprofile-use. We need to use a specific version of
+# llvm-profdata. This happens when verilator compiled with CC=clang
+# CXX=clang++. In this case, we should add LLVM_PROFDATA=llvm-profdata
+# when calling make. For GCC, this step should be skipped. Also, some
+# machines may have multiple versions of llvm-profdata. So please never
+# add default value for LLVM_PROFDATA unless we have a proper way to probe
+# the compiler and the corresponding llvm-profdata value.
+	$(LLVM_PROFDATA) merge $(EMU_PGO_DIR)/*.profraw -o $(EMU_PGO_DIR)/default.profdata
+else # ifdef LLVM_PROFDATA
+	@echo ""
+	@echo "----------------------- NOTICE BEGIN -----------------------"
+	@echo "If your verilator is compiled with LLVM, please don't forget"
+	@echo "to add LLVM_PROFDATA=llvm-profdata when calling make."
+	@echo ""
+	@echo "If your verilator is compiled with GCC, please ignore this"
+	@echo "message and NEVER adding LLVM_PROFDATA when calling make."
+	@echo "----------------------- NOTICE  END  -----------------------"
+	@echo ""
+endif # ifdef LLVM_PROFDATA
+	@echo "Building emu with PGO profile..."
+	@$(MAKE) clean_obj
+	$(TIME_CMD) $(MAKE) -s VM_PARALLEL_BUILDS=1 OPT_FAST="-O3" OPT_SLOW="-O0" PGO_CFLAGS="-fprofile-use=$(EMU_PGO_DIR)" PGO_LDFLAGS="-fprofile-use=$(EMU_PGO_DIR)" -C $(<D) -f $(<F) $(EMU_COMPILE_FILTER)
+else # ifdef PGO_WORKLOAD
+	@echo "Building emu..."
 	$(TIME_CMD) $(MAKE) -s VM_PARALLEL_BUILDS=1 OPT_FAST="-O3" OPT_SLOW="-O0" -C $(<D) -f $(<F) $(EMU_COMPILE_FILTER)
+endif # ifdef PGO_WORKLOAD
 
 $(EMU): $(EMU_MK) $(EMU_DEPS) $(EMU_HEADERS)
 ifeq ($(REMOTE),localhost)
 	@$(MAKE) build_emu_local
 else
-	ssh -tt $(REMOTE) 'export NOOP_HOME=$(NOOP_HOME); $(MAKE) -C $(NOOP_HOME)/difftest -j230 build_emu_local'
+	ssh -tt $(REMOTE) 'export NOOP_HOME=$(NOOP_HOME); $(MAKE) -C $(NOOP_HOME)/difftest -j230 build_emu_local PGO_WORKLOAD=$(PGO_WORKLOAD) LLVM_PROFDATA=$(LLVM_PROFDATA) PGO_MAX_CYCLE=$(PGO_MAX_CYCLE) PGO_EMU_ARGS=$(PGO_EMU_ARGS)'
 endif
 
 emu: $(EMU)
@@ -157,4 +198,7 @@ coverage:
 	@python3 scripts/coverage/statistics.py $(COVERAGE_DIR) > $(COVERAGE_DIR)/coverage_$(SIM_TOP).log
 	@mv $(COVERAGE_DATA) $(COVERAGE_DIR)
 
-.PHONY: build_emu_local
+clean_obj:
+	rm -f $(EMU_DIR)/*.o $(EMU_DIR)/*.gch $(EMU_DIR)/*.a $(EMU)
+
+.PHONY: build_emu_local clean_obj
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -96,6 +96,16 @@ jobs:
             source ./env.sh
             ./build/emu -b 0 -e 0 --no-diff -C 10000
 
+      - name: Basic Difftest with PGO build
+        run: |
+            cd $GITHUB_WORKSPACE/../xs-env
+            source ./env.sh
+            cd $GITHUB_WORKSPACE/../xs-env/NutShell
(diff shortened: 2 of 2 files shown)


verilator.mk
```diff
@@ -87,6 +87,10 @@ ifeq ($(EMU_COVERAGE),1)
 VEXTRA_FLAGS += --coverage-line --coverage-toggle
 endif
 
+# Profile Guided Optimization
+EMU_PGO_DIR  = $(BUILD_DIR)/pgo
```
poemonsense, line 91: What about emu-compile-pgo? We are using `emu-compile` for verilator, `simv-compile` for VCS, and `pldm-compile` for Palladium.

verilator.mk
```diff
@@ -105,9 +109,11 @@ VERILATOR_FLAGS =                   \
   --output-split 30000              \
   --output-split-cfuncs 30000       \
   -I$(RTL_DIR)                      \
-  -I$(GEN_VSRC_DIR)		    \
+  -I$(GEN_VSRC_DIR)                 \
   -CFLAGS "$(EMU_CXXFLAGS)"         \
   -LDFLAGS "$(EMU_LDFLAGS)"         \
+  -CFLAGS "\$$(PGO_CFLAGS)"         \
```
poemonsense, line 115: Will this CLAGS be appended to the original CFLAGS at line 113? I'm only asking to ensure it has been tested.

verilator.mk
```diff
@@ -138,13 +144,45 @@ EMU_COMPILE_FILTER =
 build_emu_local: $(EMU_MK)
 	@echo -e "\n[c++] Compiling C++ files..." >> $(TIMELOG)
 	@date -R | tee -a $(TIMELOG)
+ifdef PGO_WORKLOAD
+	@echo "Building PGO profile..."
+	@$(MAKE) clean_obj
+	@mkdir -p $(EMU_PGO_DIR)
+	$(TIME_CMD) $(MAKE) -s VM_PARALLEL_BUILDS=1 OPT_FAST="-O3" OPT_SLOW="-O0" PGO_CFLAGS="-fprofile-generate=$(EMU_PGO_DIR)" PGO_LDFLAGS="-fprofile-generate=$(EMU_PGO_DIR)" -C $(<D) -f $(<F) $(EMU_COMPILE_FILTER)
+	@echo "Trainning emu with PGO Workload..."
+	$(EMU) -i $(PGO_WORKLOAD) --max-cycles=$(PGO_MAX_CYCLE) 1>$(EMU_PGO_DIR)/`date +%s`.log 2>$(EMU_PGO_DIR)/`date +%s`.err
```
poemonsense, line 153: I noticed you have provided the patch for XiangShan which contains a realpath for PGO_WORKLOAD.

I'm considering adding a checker in this Makefile as well to ensure the user is passing an existed filename to PGO_WORKLOAD. Since XiangShan's Makefile is out of the control of DiffTest, we need to ensure the users of difftest passing a correct path of PGO_WORKLOAD at make-time. This implies that, as long as the user passes a legal file path (either realpath or not) with PGO_WORKLOAD, DiffTest accepts it. With this checker, external users (outside XiangShan) can connect DiffTest correctly with warnings in appropriate cases.

poemonsense (commented):
The PGO commands look good to me except for some minor changes.

I'm suggesting adding a CI test for verify the PGO process. NutShell builds quickly. I'm assuming $(MAKE) will have all variables including PGO_WORKLOAD. That to say, if you `make emu PGO_WORKLOAD=../ready-to-run/microbench.bin` would work perfectly (given the relative path starting at `difftest` is correctly pointing to a file).

verilator.mk
cyyself: > Will this CLAGS be appended to the original CFLAGS at line 113? I'm only asking to ensure it has been tested.

Yes. I have tested it.

poemonsense:
> ```diff
> diff --git a/Makefile b/Makefile
> --- a/Makefile
> +++ b/Makefile
> @@ -205,7 +205,7 @@ idea:
> 
>  # verilator simulation
>  emu: sim-verilog
> -       $(MAKE) -C ./difftest emu SIM_TOP=SimTop DESIGN_DIR=$(NOOP_HOME) NUM_CORES=$(NUM_CORES)
> +       $(MAKE) -C ./difftest emu SIM_TOP=SimTop DESIGN_DIR=$(NOOP_HOME) NUM_CORES=$(NUM_CORES) PGO_WORKLOAD=`realpath $(PGO_WORKLOAD) 2>/dev/null`
> 
>  # vcs simulation
>  simv: sim-verilog
> ```

Is it better to move this realpath to separate lines in Makefile?

```
ifneq ($(PGO_WORKLOAD),)
// update PGO_WORLOAD here with realpath
endif
```

Do we really need to pass the PGO_WORKLOAD explicitly? I'm assuming all variables are perserved by $(MAKE)

verilator.mk
poemonsense: I just noticed line 113 is "$(EMU_CXXFLAGS)" and line 115 is "\$$(PGO_CFLAGS)"

Are `$` and `\$$` same?

verilator.mk
```diff
@@ -138,13 +144,45 @@ EMU_COMPILE_FILTER =
...
+# directly be used in -fprofile-use. We need to use a specific version of
+# llvm-profdata. This happens when verilator compiled with CC=clang
+# CXX=clang++. In this case, we should add LLVM_PROFDATA=llvm-profdata
+# when calling make. For GCC, this step should be skipped. So please
+# never add default value for LLVM_PROFDATA.
+	$(LLVM_PROFDATA) merge $(EMU_PGO_DIR)/*.profraw -o $(EMU_PGO_DIR)/default.profdata
+else # ifdef LLVM_PROFDATA
+	@echo ""
+	@echo "----------------------- NOTICE BEGIN -----------------------"
+	@echo "If your varilator is compiled with LLVM, please don't forget"
```
eastonman, line 165: a typo here for verilator

verilator.mk
```diff
@@ -138,13 +144,45 @@ EMU_COMPILE_FILTER =
...
+	@echo "to add LLVM_PROFDATA=llvm-profdata when calling make."
+	@echo ""
+	@echo "If your varilator is compiled with GCC, please ignore this"
```
eastonman, line 168: here too

eastonman (changes requested):

eastonman:
Do you have some simulation performance stats you can share?
And how much time does max-cycle 2M take on an ordinary EPYC 2nd or 3rd gen machine?

verilator.mk
cyyself: No.

1.`$$` is used in Makefile to produce `$`: https://www.gnu.org/software/make/manual/html_node/Variables-in-Recipes.html
2. Then, `\$` is used in the shell to pass the string rather than the variable name.

verilator.mk
cyyself: That sounds like a good idea. I will fix it tomorrow.

cyyself:
> > ```diff
> > diff --git a/Makefile b/Makefile
> > --- a/Makefile
> > +++ b/Makefile
> > @@ -205,7 +205,7 @@ idea:
> > 
> >  # verilator simulation
> >  emu: sim-verilog
> > -       $(MAKE) -C ./difftest emu SIM_TOP=SimTop DESIGN_DIR=$(NOOP_HOME) NUM_CORES=$(NUM_CORES)
> > +       $(MAKE) -C ./difftest emu SIM_TOP=SimTop DESIGN_DIR=$(NOOP_HOME) NUM_CORES=$(NUM_CORES) PGO_WORKLOAD=`realpath $(PGO_WORKLOAD) 2>/dev/null`
> > 
> >  # vcs simulation
> >  simv: sim-verilog
> > ```
> 
> Is it better to move this realpath to separate lines in Makefile?
> 
> ```
> ifneq ($(PGO_WORKLOAD),)
> // update PGO_WORLOAD here with realpath
> endif
> ```
> 
> Do we really need to pass the PGO_WORKLOAD explicitly? I'm assuming all variables are perserved by $(MAKE)

It's just for adding the real path and passed to another Makefile.

cyyself:
> Do you have some simulation performance stats you can share? And how much time does max-cycle 2M take on an ordinary EPYC 2nd or 3rd gen machine?

I haven't tested it on the AMD EPYC Server. On my 13900k machine with 16 threads, with coremark trained PGO and coremark emu benchmark, it gets about 2x faster, which finished the 2 iter coremark in 46s, original emu takes about 120s. This result is about 1x faster than non-PGO emu on 7950X3D, which takes 76s to finish 2 iter coremark. However, I think we may need a better workload as a training workload that uses MMU and RVV.

verilator.mk
cyyself: > What about emu-compile-pgo? We are using `emu-compile` for verilator, `simv-compile` for VCS, and `pldm-compile` for Palladium.

After thinking for a while, I think keeping `emu-compile` might be better. Since the C compiler does the PGO, the C code generated by the verilator can be reused. Also, compiling Verilog to C takes time on a single thread, but compiling C code to binary is very fast. This will also reduce users' switching time from normal emu to PGO emu.

verilator.mk
cyyself: I have added `stat` in line 149, and I think a checker like this is OK.

verilator.mk
poemonsense: I see

verilator.mk
poemonsense: `BUILD_DIR` is `build`. `EMU_DIR` is `build/emu-compile`.

I suggest using `build/emu-compile-pgo` for `EMU_PGO_DIR` instead of `build/pgo`. You may also use `EMU_DIR/pgo`.

poemonsense:
> The PGO commands look good to me except for some minor changes.
> 
> I'm suggesting adding a CI test for verify the PGO process. NutShell builds quickly. I'm assuming $(MAKE) will have all variables including PGO_WORKLOAD. That to say, if you `make emu PGO_WORKLOAD=../ready-to-run/microbench.bin` would work perfectly (given the relative path starting at `difftest` is correctly pointing to a file).


There are two unresolved comments

1) It's better to add a CI test.

2) Better to use `build/emu-compile-pgo` or `build/emu-compile/pgo` instead of `build/pgo` which confuses the user.

verilator.mk
cyyself: OK. Solved.

poemonsense (dismissed):
Approving it now.

Please let me know when it's ready for merging. CI may be added in later PRs.

verilator.mk
```diff
@@ -138,13 +144,48 @@ EMU_COMPILE_FILTER =
 build_emu_local: $(EMU_MK)
 	@echo -e "\n[c++] Compiling C++ files..." >> $(TIMELOG)
 	@date -R | tee -a $(TIMELOG)
+ifdef PGO_WORKLOAD
+	@echo "Building PGO profile..."
+	@stat $(PGO_WORKLOAD) > /dev/null
+	@$(MAKE) clean_obj
+	@mkdir -p $(EMU_PGO_DIR)
+	$(TIME_CMD) $(MAKE) -s VM_PARALLEL_BUILDS=1 OPT_FAST="-O3" OPT_SLOW="-O0" PGO_CFLAGS="-fprofile-generate=$(EMU_PGO_DIR)" PGO_LDFLAGS="-fprofile-generate=$(EMU_PGO_DIR)" -C $(<D) -f $(<F) $(EMU_COMPILE_FILTER)
+	@echo "Trainning emu with PGO Workload..."
```
poemonsense, line 153: Typo: training

verilator.mk
```diff
@@ -87,6 +87,10 @@ ifeq ($(EMU_COVERAGE),1)
 VEXTRA_FLAGS += --coverage-line --coverage-toggle
 endif
 
+# Profile Guided Optimization
+EMU_PGO_DIR  = $(EMU_DIR)/pgo
```
poemonsense, line 91: I noticed $(EMU_DIR) is defined in line 91. Do we need to move this line after line 91?

cyyself: Oh. It's my fault. Fixed.

cyyself:
Maybe I should add a CI based on NutShell. But I'm not sure I have time to do that recently. I'm figuring out the problem caused by a recent update on the XiangShan master branch, which can not finish C compile with -fprofile-generate and O3 or -Os on llvm-18 in 30 minutes, which makes PGO unusable. But with -O0, there is a lot of performance degradation.

poemonsense:
> Maybe I should add a CI based on NutShell. But I'm not sure I have time to do that recently. I'm figuring out the problem caused by a recent update on the XiangShan master branch, which can not finish C compile with -fprofile-generate and (O3 or -Os) on llvm-18 in 30 minutes, which makes PGO unusable. But with -O0, there is a lot of performance degradation.

We can leave this PR unmerged until these issues fixed.

cyyself:
> It's a very good work. Thanks for the contribution. There are two unresolved comments
> 
> 1. It's better to add a CI test.

CI added now.

> 2. Better to use `build/emu-compile-pgo` or `build/emu-compile/pgo` instead of `build/pgo` which confuses the user.

I use `build/emu-compile/pgo` now.

cyyself:
And please don't merge in squash mode; I have some important messages in each commit message.

cyyself:
> > Maybe I should add a CI based on NutShell. But I'm not sure I have time to do that recently. I'm figuring out the problem caused by a recent update on the XiangShan master branch, which can not finish C compile with -fprofile-generate and (O3 or -Os) on llvm-18 in 30 minutes, which makes PGO unusable. But with -O0, there is a lot of performance degradation.
> 
> We can leave this PR unmerged until these issues fixed.

I think we can merge it first and then add a parameter in Makefile to switch the optimization level for difftest without rerunning the verilator. Now clang-18 is broken on the verilator-generated XiangShan RTL. So I can’t provide a fair comparison with different optimization levels. But one thing I can tell is the instruction count on 8 threads coremark is nearly the same on GCC-14 -O3 and LLVM-18 -O3. But if we switched to -O1 on GCC-14, the instruction count on GCC-14 would be 3x compared to -O3. So providing a switch might be a good choice if we ran into a compile-time issue or emu performance issue.

poemonsense (approved):

PR #394, opened 2024-06-12
vcs: difftest checks on negedge

Tang-Haojin:
Difftest is essentially a producer-consumer framework, where various DPI-C functions in SimTop.v act as producers, and functions like simv_nstep serve as consumers. Currently, in the difftest framework based on vcs / tb_top, both producers and consumers operate on the rising edge, leading to a race and hazard issue, namely: before all the DPI-C functions of the producer have executed, the consumer starts comparing with incomplete content, thereby triggering errors. This issue has already occurred in vcs.

As shown in the figure below, when the commit in the ROB is enabled, the mstatus in the CSR also changes synchronously.

However, judging from the printed content (printf in the DPI-C functions in ROB commit and CSR), the actual execution order is: the commit DPI-C function in the ROB -> the difftest comparison process (comparison fails, the producer function of the CSR has not yet been executed) -> (the DPI-C function of the CSR may be executed at this time).

It appears that a simple solution is for the producer to produce on the rising edge, while the consumer consumes on the falling edge.

Feel free to discuss whether this solution is reasonable, whether the modifications in the PR are complete, and whether there are better solutions.

### DIFF
--- a/src/test/vsrc/vcs/DeferredControl.v
+++ b/src/test/vsrc/vcs/DeferredControl.v
@@ -35,7 +35,7 @@ initial $ixc_ctrl("gfifo", "simv_nstep");
 `endif // PALLADIUM
 `endif // CONFIG_DIFFTEST_NONBLOCK
 
-always @(posedge clock) begin
+  always @(negedge clock) begin
   if (reset) begin
     simv_result <= 8'b0;
   end
--- a/src/test/vsrc/vcs/DeferredControl.v
+++ b/src/test/vsrc/vcs/DeferredControl.v
@@ -35,7 +35,7 @@ initial $ixc_ctrl("gfifo", "simv_nstep");
 `endif // PALLADIUM
 `endif // CONFIG_DIFFTEST_NONBLOCK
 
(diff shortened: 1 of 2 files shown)


poemonsense:
Let @klin02 have a review. 

I remembered he told me about similar race issues before.

klin02:
In Gateway.scala, we delay step one cycle to make sure consume after produce. And we can set ping-pong buffer for consumer and producer by enable difftest-config Z.

Just try pass “--difftest-config Z” or EZ to mill. See detail in Makefile of XS.

klin02:
Related discussion #283

Tang-Haojin:
Seems that `--difftest-config Z` can resolve this issue. I think it may be set as default option when we use `tb_top`.

poemonsense:
> Seems that `--difftest-config Z` can resolve this issue. I think it may be set as default option when we use `tb_top`.

Are we required to built different Verilog for VCS and Verilator? @klin02 

If so, it looks strange. Are there any alternative approaches to fix the issue?

PR #400, opened 2024-06-19
difftest: add out workload cpi with file

xiaokamikami:
When  enabled ”OUTPUT_CPI_TO_FILE“ , the [id,CPI] will be output to a temporary file at the end of the simulation, This information is mainly used to support the generation of multi-core checkpoints of the qemu linkage EMU simulation model, which helps to improve the sampling progress and sampling speed。

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,9 +120,14 @@ SIM_CXXFLAGS += -DWITH_DRAMSIM3 -DDRAMSIM3_CONFIG=\\\"$(DRAMSIM3_HOME)/configs/X
 SIM_LDFLAGS  += $(DRAMSIM3_HOME)/build/libdramsim3.a
 endif
 
-# Out ipc info on txt file, mainly applied to support qemu multi-core sampled data output
-ifeq ($(OUTPUT_CPI_TO_FIFO),1)
-SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FIFO
+# out ipc info on temporary txt file, mainly applied to support qemu multi-core sampled data
+ifeq ($(OUTPUT_CPI_TO_FILE),1)
+SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE
+ifneq ($(origin OUTPUT_CPI_TO_FILE_DIR), undefined)
+SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE_DIR=$(OUTPUT_CPI_TO_FILE_DIR)
+else
+SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE_DIR=\\\"$(shell pwd)/emu_to_cpi_file.txt\\\"
+endif
 endif
 
 ifeq ($(PMEM_CHECK),1)
--- a/Makefile
+++ b/Makefile
@@ -125,8 +125,6 @@ ifeq ($(OUTPUT_CPI_TO_FILE),1)
 SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE
 ifneq ($(origin OUTPUT_CPI_TO_FILE_DIR), undefined)
 SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE_DIR=$(OUTPUT_CPI_TO_FILE_DIR)
-else
-SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE_DIR=\\\"$(shell pwd)/emu_to_cpi_file.txt\\\"
 endif
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -121,11 +121,8 @@ SIM_LDFLAGS  += $(DRAMSIM3_HOME)/build/libdramsim3.a
 endif
 
 # out ipc info on temporary txt file, mainly applied to support qemu multi-core sampled data
-ifeq ($(OUTPUT_CPI_TO_FILE),1)
-SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE
-ifneq ($(origin OUTPUT_CPI_TO_FILE_DIR), undefined)
-SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE_DIR=$(OUTPUT_CPI_TO_FILE_DIR)
-endif
+ifneq ($(OUTPUT_CPI_TO_FILE),)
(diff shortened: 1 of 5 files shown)


src/test/csrc/difftest/difftest.cpp
```diff
@@ -1188,6 +1188,12 @@ void Difftest::display_stats() {
   double ipc = (double)instrCnt / cycleCnt;
   eprintf(ANSI_COLOR_MAGENTA "Core-%d instrCnt = %'" PRIu64 ", cycleCnt = %'" PRIu64 ", IPC = %lf\n" ANSI_COLOR_RESET,
           this->id, instrCnt, cycleCnt, ipc);
+#ifdef WITH_QEMUCKPT
```
poemonsense, line 1191: The macro name should strictly match the code.

Here the code is for output of CPI. Then the macro should be named as OUTPUT_CPI_TO_FILE or other names with similar meaning.

The current code seems to have nothing to do with QEMUCKPT.

src/test/csrc/difftest/difftest.cpp
```diff
@@ -1188,6 +1188,12 @@ void Difftest::display_stats() {
   double ipc = (double)instrCnt / cycleCnt;
   eprintf(ANSI_COLOR_MAGENTA "Core-%d instrCnt = %'" PRIu64 ", cycleCnt = %'" PRIu64 ", IPC = %lf\n" ANSI_COLOR_RESET,
           this->id, instrCnt, cycleCnt, ipc);
+#ifdef OUTPUT_CPI_TO_FILE
```
klin02, line 1191: Please clarify IPC and CPI. Macros seems inconsistent with variable.

palladium.mk
```diff
@@ -30,6 +30,10 @@ endif
 ifeq ($(WITH_DRAMSIM3),1)
 PLDM_MACRO_FLAGS  	+= +define+WITH_DRAMSIM3
 endif
+# Out ipc info on txt file, mainly applied to support qemu multi-core sampled data output
+ifeq ($(OUTPUT_CPI_TO_FILE),1)
+PLDM_MACRO_FLAGS  	+= +define+OUTPUT_CPI_TO_FILE
+endif
```
klin02, line 36: Seems the func also supports VCS? If so, shall we also add it in vcs.mk

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -35,6 +35,9 @@
 #include "perf.h"
 #endif // CONFIG_DIFFTEST_PERFCNT
 
+#ifdef OUTPUT_CPI_TO_FILE
+char out_ipc_info_file[32] = "./emu_to_cpi_file.txt";
```
klin02, line 39: Maybe the file name should not be fixed in code. Pass it by Macro or DPIC?

xiaokamikami: This file exists as an auto-generated temporary file, so you shouldn't need to add any more arguments to the run command

klin02:
display_stats is a printf-like function, we should not use return val.

Wrap related logic into a function, rather than in simv_step.

Makefile
```diff
@@ -120,6 +120,16 @@ SIM_CXXFLAGS += -DWITH_DRAMSIM3 -DDRAMSIM3_CONFIG=\\\"$(DRAMSIM3_HOME)/configs/X
 SIM_LDFLAGS  += $(DRAMSIM3_HOME)/build/libdramsim3.a
 endif
 
+# out ipc info on temporary txt file, mainly applied to support qemu multi-core sampled data
+ifeq ($(OUTPUT_CPI_TO_FILE),1)
+SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE
+ifneq ($(origin OUTPUT_CPI_TO_FILE_DIR), undefined)
+SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE_DIR=$(OUTPUT_CPI_TO_FILE_DIR)
+else
+SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE_DIR=\\\"$(shell pwd)/emu_to_cpi_file.txt\\\"
```
klin02, line 129: We can pass it by Macro. No need to write a fixed name here, then seems a single Macro is enough.

Makefile
```diff
@@ -120,6 +120,14 @@ SIM_CXXFLAGS += -DWITH_DRAMSIM3 -DDRAMSIM3_CONFIG=\\\"$(DRAMSIM3_HOME)/configs/X
 SIM_LDFLAGS  += $(DRAMSIM3_HOME)/build/libdramsim3.a
 endif
 
+# out ipc info on temporary txt file, mainly applied to support qemu multi-core sampled data
+ifeq ($(OUTPUT_CPI_TO_FILE),1)
```
poemonsense, line 124: Macros can be simplified. Only `OUTPUT_CPI_TO_FILE` is required. When it is defined, it should refer to a file path.

```Makefile
ifneq ($(OUTPUT_CPI_TO_FILE),)
SIM_CXXFLAGS += -DOUTPUT_CPI_TO_FILE=$(OUTPUT_CPI_TO_FILE)
endif
```

Similar changes apply to C++ as well. There's no need to use two different macros.

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -164,6 +165,21 @@ extern "C" uint8_t simv_init() {
   return 0;
 }
 
+#ifdef OUTPUT_CPI_TO_FILE
+int output_cpi_to_file() {
+  FILE *d2q_fifo = fopen(OUTPUT_CPI_TO_FILE_DIR, "w+");
```
poemonsense, line 170: fopen should have its fclose.

xiaokamikami: ok

Makefile
xiaokamikami: ok

poemonsense (approved):
Look good to me. I'm wondering why we make it a compile-time macro instead of a run-time command line argument?

xiaokamikami:
> Look good to me. I'm wondering why we make it a compile-time macro instead of a run-time command line argument?

If set to run parameter, when I don't want to enable this function, I need to recompile verilog. This function is only related to the software side, so I don't need to set it from the run command in the simulation

PR #401, opened 2024-06-19
SimJTAG: add `--remote-jtag-port` to specify remote bitbang port

Tang-Haojin:


src/test/csrc/common/SimJTAG.cpp
```diff
@@ -19,7 +20,7 @@ extern "C" int jtag_tick(unsigned char *jtag_TCK, unsigned char *jtag_TMS, unsig
     return 0;
   if (!jtag) {
     // TODO: Pass in real port number
```
poemonsense, line 22: Let's remove this TODO and add a separate function `jtag_init`. This function can be reused here in SimJTAG.cpp as well as emu.cpp.

```c++
void jtag_init() {
  jtag = new remote_bitbang_t(remote_jtag_port);
}

// in SimJTAG.cpp
if(!jtag) {
  jtag_init();
};

// in emu.cpp
if (args.enable_jtag) {
  remote_jtag_port = args.remote_jtag_port;
  jtag_init();
}
```

src/test/csrc/verilator/emu.cpp
```diff
@@ -134,6 +135,7 @@ inline EmuArgs parse_args(int argc, const char *argv[]) {
     { "no-diff",           0, NULL,  0  },
     { "enable-fork",       0, NULL,  0  },
     { "enable-jtag",       0, NULL,  0  },
+    { "remote-jtag-port",  1, NULL,  0  },
```
poemonsense, line 138: In the `--help` message, it is better to put it after `--enable-jtag`. 

However, I think we do not need to put this option here. We can append this argument at the end of the argument list (struct). Then we do not need to change the indices for prior arguments, and there will be less lines of code changes. This order change will not affect the functionalities.

poemonsense (commented):
Propose some style changes

src/test/csrc/verilator/emu.h
```diff
@@ -37,6 +37,7 @@
 #endif
 
 struct EmuArgs {
+  uint16_t remote_jtag_port = 23334;
```
poemonsense, line 40: We can leave the default value only at SimJTAG.cpp.

We also do not need to add the argument in EmuArgs. We can set remote_jtag_port when parsing the arguments.

```c++
case SOME_NUMBER: remote_jtag_port  = atoll_strict(optarg, "remote-jtag-port"); continue;
```

src/test/csrc/verilator/emu.cpp
Tang-Haojin: Fixed.

src/test/csrc/verilator/emu.h
Tang-Haojin: Fixed. Also remove redundant `enable_jtag`.

src/test/csrc/common/SimJTAG.cpp
Tang-Haojin: Fixed.

poemonsense (approved):

PR #403, opened 2024-06-20
CommitData: preprocess data in Hardware, submit with InstrCommit

klin02:
We use physical Register WriteBack for loadEvent and MMIO access, and record commit Data in Commit Instr Trace. As there are multiple DUT buffer in software side, WriteBacks transferred and used may not in same cycle, so we buffer corresponding wb data until instrCommit.

As writeBacks only used in get_commit_data, and comes from int/vec/fp, we preprocess commitData in hardware side, and submit along with InstrCommit. This change will fix error of fpWriteBacks, and also help simplify logic of Load Squash.

### DIFF
--- a/src/main/scala/Bundles.scala
+++ b/src/main/scala/Bundles.scala
@@ -89,6 +89,11 @@ class InstrCommit(val numPhyRegs: Int = 32) extends DifftestBaseBundle with HasV
   }
 }
 
+// Instantiate inside DiffTest, work for get_commit_data specially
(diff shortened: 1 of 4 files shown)


klin02:
After this change, Dual-core difftest reach 14.88 KHz with config "ZEN". Seems still some bug of Squash feature, likely related to loadEvent. I will try to fix that based on this change.

poemonsense (approved):

PR #405, opened 2024-06-24
SquashQueue: queue unsquash event and check against stamp

klin02:
Some unsquash event, such as LoadEvent and StoreEvent, will seriously hinder squash of related diffstate. We submit such event independently and queue them in software side with SquashStamp.

There queued event will be checked when num_commit meets stamp, resuming check order of different squashed event.

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -257,6 +257,12 @@ int Difftest::step() {
   store_event_record();
 #endif
 
+#ifdef CONFIG_DIFFTEST_SQUASH
+#ifdef CONFIG_DIFFTEST_LOADEVENT
+  load_event_record();
+#endif // CONFIG_DIFFTEST_LOADEVENT
+#endif // CONFIG_DIFFTEST_SQUASH
+
 #ifdef DEBUG_GOLDENMEM
   if (do_golden_memory_update()) {
     return 1;
@@ -317,9 +323,12 @@ int Difftest::step() {
         if (do_instr_commit(i)) {
           return 1;
         }
+#ifndef CONFIG_DIFFTEST_SQUASH
+        do_load_check(i);
         if (do_store_check()) {
           return 1;
         }
+#endif // CONFIG_DIFFTEST_SQUASH
         dut->commit[i].valid = 0;
         num_commit += 1 + dut->commit[i].nFused;
       }
@@ -497,11 +506,17 @@ int Difftest::do_instr_commit(int i) {
     return 0;
   }
 
-  // single step exec
-  proxy->ref_exec(1);
+  // Default: single step exec
   // when there's a fused instruction, let proxy execute more instructions.
-  for (int j = 0; j < dut->commit[i].nFused; j++) {
+  for (int j = 0; j < dut->commit[i].nFused + 1; j++) {
     proxy->ref_exec(1);
+#ifdef CONFIG_DIFFTEST_SQUASH
+    commit_stamp = (commit_stamp + 1) % CONFIG_DIFFTEST_SQUASH_STAMPSIZE;
+    do_load_check(i);
+    if (do_store_check()) {
+      return 1;
+    }
+#endif // CONFIG_DIFFTEST_SQUASH
   }
 
   return 0;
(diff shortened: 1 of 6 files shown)


klin02:
Now dual-core difftest speed is nearly 23 KHz. The speed is mainly blocked by globalEnable, which means we submit diffState uncontrolled by valid when there are other state valid in the same cycle.
We will try to deal with that in another PR.

klin02:
We test this feature in longer workload (40M instrs) to reduces block of DPIC Memory read. Then the speed is nearly 140 KHz. And ideal speed ( No difftest, only DPIC Mem) is 420 KHz.

poemonsense (approved):
Generally it is not usual to let the checkers depend on how DiffTest organizes the data. But I'm approving it to allow first merging it.

Ideally, the checkers should never know how the data is transferred. What the checkers see should be a C++ class for the overall DUT state. If squash affects the definition of the state class for cosim, we should fix it in another layer instead changing how cosim works. Ideally the abstractions should look like: Hardware interfaces -> DiffTest Verilog side interfaces -> DiffTest C++ side interfaces (for data transmission) -> DiffTest C++ side interfaces (for cosim).

Previously we are generating some constructors in C++. For example, at https://github.com/OpenXiangShan/difftest/blob/master/src/main/scala/DPIC.scala#L356-L382. I think similar methods would be more suitable for this feature.

We should always distinguish the following stages clearly: 1) DiffTest interfaces for DUT; 2) DiffTest Transforms; 3) Hardware-Software communication methods; 4) cosim logic in the software side. Each step should take care of its own inputs/outputs. Different stages should have a clear interface between each other

src/test/csrc/difftest/difftest.cpp
```diff
@@ -542,45 +557,58 @@ void Difftest::do_load_check(int i) {
...
+    auto refRegPtr = proxy->arch_reg(load_event.wdest, load_event.fpwen);
+    auto commitData = load_event.commitData;
+#else
+    auto load_event = dut->load[i];
+    if (!load_event.valid)
       return;
     bool regWen =
         ((dut->commit[i].rfwen && dut->commit[i].wdest != 0) || dut->commit[i].fpwen) && !dut->commit[i].vecwen;
-    if (dut->load[i].isLoad || dut->load[i].isAtomic) {
+    auto refRegPtr = proxy->arch_reg(dut->commit[i].wdest, dut->commit[i].fpwen) auto commitData = get_commit_data(i);
```
poemonsense, line 575: Here's a missing `;` before `auto commitData`, which causes errors in https://github.com/OpenXiangShan/XiangShan/actions/runs/9772624822/job/26977314449?pr=3133, https://github.com/OpenXiangShan/XiangShan/pull/3133.

@klin02 Why does not CI find it? Please help fix it

klin02: Now Difftest CI only check NutShell as DUT, which didn't check LoadEvent. Maybe we can add Nightly CI for XiangShan?

src/test/csrc/difftest/difftest.cpp
```diff
@@ -497,11 +506,17 @@ int Difftest::do_instr_commit(int i) {
...
-  // single step exec
-  proxy->ref_exec(1);
+  // Default: single step exec
   // when there's a fused instruction, let proxy execute more instructions.
-  for (int j = 0; j < dut->commit[i].nFused; j++) {
+  for (int j = 0; j < dut->commit[i].nFused + 1; j++) {
     proxy->ref_exec(1);
+#ifdef CONFIG_DIFFTEST_SQUASH
+    commit_stamp = (commit_stamp + 1) % CONFIG_DIFFTEST_SQUASH_STAMPSIZE;
+    do_load_check(i);
```
poemonsense, line 515: Why is the `i` used here inside a `j` loop?

src/test/csrc/difftest/difftest.cpp
```diff
@@ -497,11 +506,17 @@ int Difftest::do_instr_commit(int i) {
...
+    if (do_store_check()) {
+      return 1;
```
poemonsense, line 517: Also, these `load_check` and `store_check` seem duplicated with the ones in `do_instr_commit`. Only commit_stamp is changed, others look the same

PR #406, opened 2024-06-26
Validate: append valid to DifftestBundle

klin02:
Previous we transmit all DiffState without hasValid when some DiffState valid in same cycle. However, when Squash enabled, some State without hasValid can be validated with special condition. As Submit times of State with SquashQueue is much greater than others, validated will greatly reduce submit times of Stata without hasValid.

Previous we also use do_squash and squashDependency to filter invalid CSR value. This change rename squashDependency to updateDependency, and only validate CSR when commit and event. So we remove do_squash.

As Trapevent will also be checked for instrs Exceed, we transmit it when validated, rather than just hasTrap or hasWFI.

### DIFF
--- a/src/main/scala/Batch.scala
+++ b/src/main/scala/Batch.scala
@@ -55,8 +55,8 @@ class BatchInfo extends Bundle {
 object Batch {
   private val template = ListBuffer.empty[DifftestBundle]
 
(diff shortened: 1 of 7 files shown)


poemonsense (approved):

klin02:
Have been tested on Palladium. Dual-core speed reachs 140 Khz, as expected.

PR #415, opened 2024-07-09
util: add support for synthesizable assertions

poemonsense:
We add a VerificationExtractor transform with an annotation named VerificationExtractorSink to support synthesizable assertions.

First, the user should annotate a sink Bool variable to let the transform know that it should do wiring for all assertions. Second, the transform collects all assertions (firrtl.ir.Verification) from the circuit and adds SourceAnnotation for wiring. Third, in the sink module, the transform creates Cat and orR nodes to compute a Bool value for the annotated assertion sink.

The user should 1) annotate the sink Bool using the function VerificationExtractor.sink, and 2) add VerificationExtractor to the build dependecy as one of the transforms.

This transform does not support Chisel 3.x now and in the future. To allow compiling the code using Chisel 3.x, we use Scala reflection to create DefRegisterWithReset nodes.

It's worth noting that, this transform only creates a 1-bit value for detecting the assertions. It does not support any debugging methodologies to expose the 1-bit value to the simulation environment. The debugging method should depend on the simulation platform.

### DIFF
--- a/src/main/scala/util/VerificationExtractor.scala
+++ b/src/main/scala/util/VerificationExtractor.scala
@@ -1,0 +1,168 @@
+/***************************************************************************************
+ * Copyright (c) 2020-2024 Institute of Computing Technology, Chinese Academy of Sciences
+ *
+ * DiffTest is licensed under Mulan PSL v2.
+ * You can use this software according to the terms and conditions of the Mulan PSL v2.
+ * You may obtain a copy of Mulan PSL v2 at:
+ *          http://license.coscl.org.cn/MulanPSL2
+ *
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND,
+ * EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+ *
+ * See the Mulan PSL v2 for more details.
+ ***************************************************************************************/
+
+package difftest.util
+
+import firrtl._
+import firrtl.annotations._
+import firrtl.ir._
+import firrtl.options.Phase
+import firrtl.passes.wiring.{SinkAnnotation, SourceAnnotation}
+import firrtl.stage.FirrtlCircuitAnnotation
+
+// This is the main user interface for defining the Verification sink. Should be called only once.
+// If any sink is defined, the following VerificationExtractor transform will perform the wiring.
+object VerificationExtractor {
+  def sink(cond: chisel3.Bool): Unit = {
+    chisel3.experimental.annotate(new chisel3.experimental.ChiselAnnotation {
+      override def toFirrtl: Annotation = VerificationExtractorSink(cond.toTarget)
+    })
+  }
+}
+
+// This transform converts firrtl.ir.Verification to Wiring when the sink is defined via annotation.
+class VerificationExtractor extends Phase {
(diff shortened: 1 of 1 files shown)


poemonsense:
I don't know how to add a CI test for this. However, I've verified its functionalities with tests on NutShell.

Patch:

```diff
diff --git a/src/main/scala/sim/NutShellSim.scala b/src/main/scala/sim/NutShellSim.scala
index ce4edfe..14167fc 100644
--- a/src/main/scala/sim/NutShellSim.scala
+++ b/src/main/scala/sim/NutShellSim.scala
@@ -20,6 +20,7 @@ import bus.axi4._
 import chisel3._
 import device.AXI4RAM
 import difftest._
+import difftest.util.VerificationExtractor
 import nutcore.NutCoreConfig
 import system._
 
@@ -43,4 +44,8 @@ class SimTop extends Module {
 
   val difftest = DifftestModule.finish("nutshell")
   difftest.uart <> mmio.io.uart
+
+  val assertion = WireInit(false.B)
+  dontTouch(assertion)
+  VerificationExtractor.sink(assertion)
 }

diff --git a/src/test/scala/TopMain.scala b/src/test/scala/TopMain.scala
index 8e67f9f..0026bfd 100644
--- a/src/test/scala/TopMain.scala
+++ b/src/test/scala/TopMain.scala
@@ -24,6 +24,9 @@ import difftest.DifftestModule
 import nutcore.NutCoreConfig
 import sim.SimTop
 import system.NutShell
+import chisel3.experimental.ChiselAnnotation
+import firrtl.AnnotationSeq
+import firrtl.options.{Dependency, PhaseManager}
 
 class Top extends Module {
   val io = IO(new Bundle{})
@@ -36,6 +39,31 @@ class Top extends Module {
   dontTouch(vga.io)
 }
 
+class GeneratorStage extends ChiselStage {
+  // Keep this same as ChiselStage.run except for Dependency[difftest.util.VerificationExtractor]
+  override def run(annotations: AnnotationSeq): AnnotationSeq = {
+    val pm = new PhaseManager(
+      targets = Seq(
+        Dependency[chisel3.stage.phases.AddImplicitOutputFile],
+        Dependency[chisel3.stage.phases.AddImplicitOutputAnnotationFile],
+        Dependency[chisel3.stage.phases.MaybeAspectPhase],
+        Dependency[chisel3.stage.phases.AddSerializationAnnotations],
+        Dependency[chisel3.stage.phases.Convert],
+        Dependency[difftest.util.VerificationExtractor],
+        Dependency[chisel3.stage.phases.AddDedupGroupAnnotations],
+        Dependency[chisel3.stage.phases.MaybeInjectingPhase],
+        Dependency[circt.stage.phases.AddImplicitOutputFile],
+        Dependency[circt.stage.phases.CIRCT],
+      ),
+      currentState = Seq(
+        Dependency[firrtl.stage.phases.AddDefaults],
+        Dependency[firrtl.stage.phases.Checks],
+      ),
+    )
+    pm.transform(annotations)
+  }
+}
+
 object TopMain extends App {
   def parseArgs(info: String, args: Array[String]): String = {
     var target = ""
@@ -75,7 +103,7 @@ object TopMain extends App {
   var exe_args = newArgs.filter{
     value => value.forall(char => char!='=')
   }
-  (new ChiselStage).execute(newArgs, Seq(generator)
+  (new GeneratorStage).execute(newArgs, Seq(generator)
     :+ CIRCTTargetAnnotation(CIRCTTarget.SystemVerilog)
     :+ FirtoolOption("--disable-annotation-unknown")
   )
```

Results: assertion registers are generated at different modules and wiring to the top SimTop.

PR #417, opened 2024-07-09
emu: fix cycle conflict in full-wave dumping

Kumonda221-CrO3:
Three cycles of vcd dumping confliction after reset with **```--dump-wave-full```** in current implementation.


poemonsense (approved):
It seems strange we are having both cycle and cycles in the same class.

@klin02 please help take a look when you refactor the difftest timeout checker. It seems we need to rethink and define clearly the interfaces between simulation DUT and difftest.

PR #419, opened 2024-07-12
Wiring: support hierarchical source/sink without extra IOs 

klin02:
By default we use BoringUtils.addSink/Source to collect Difftest Signals from DUT to DifftestEndpoint, which will introduce extra IOs to each Layer of DUT Module.

By BoringUtils.tapAndRead, we can use Difftest Signals with Cross Module Reference(like SimTop.xxx.yyy) instead of explicit IO. This feature enabled by GatewayConfig.hierarchicalWiring, and requires addSource before addSink.

Since BoringUtils.tapAndRead does not exists in older version of Chisel, we use Scala reflection features to load it dynamically.

### DIFF
--- a/src/main/scala/Gateway.scala
+++ b/src/main/scala/Gateway.scala
@@ -38,6 +38,7 @@ case class GatewayConfig(
   hasInternalStep: Boolean = false,
   isNonBlock: Boolean = false,
   hasBuiltInPerf: Boolean = false,
+  hierarchicalWiring: Boolean = false,
 ) {
   def dutZoneSize: Int = if (hasDutZone) 2 else 1
   def dutZoneWidth: Int = log2Ceil(dutZoneSize)
@@ -108,6 +109,7 @@ object Gateway {
       case 'I' => config = config.copy(hasInternalStep = true)
       case 'N' => config = config.copy(isNonBlock = true)
       case 'P' => config = config.copy(hasBuiltInPerf = true)
+      case 'H' => config = config.copy(hierarchicalWiring = true)
       case x   => println(s"Unknown Gateway Config $x")
     }
     config.check()
@@ -117,7 +119,7 @@ object Gateway {
     val bundle = WireInit(0.U.asTypeOf(gen))
     if (config.needEndpoint) {
       val packed = WireInit(bundle.asUInt)
-      DifftestWiring.addSource(packed, s"gateway_${instances.length}")
+      DifftestWiring.addSource(packed, s"gateway_${instances.length}", config.hierarchicalWiring)
     } else {
       val control = WireInit(0.U.asTypeOf(new GatewaySinkControl(config)))
       control.enable := true.B
@@ -129,7 +131,13 @@ object Gateway {
 
   def collect(): GatewayResult = {
     val sink = if (config.needEndpoint) {
-      val endpoint = Module(new GatewayEndpoint(instances.toSeq, config))
+      val signals = instances.toSeq
+      val packed = WireInit(0.U.asTypeOf(MixedVec(signals.map(gen => UInt(gen.getWidth.W)))))
+      for ((data, idx) <- packed.zipWithIndex) {
+        DifftestWiring.addSink(data, s"gateway_$idx", config.hierarchicalWiring)
+      }
+      val endpoint = Module(new GatewayEndpoint(signals, config))
+      endpoint.in := packed
       GatewayResult(
         instances = endpoint.instances,
         structPacked = Some(config.isBatch),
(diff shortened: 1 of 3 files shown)


src/main/scala/util/DataMirror.scala
```diff
@@ -33,5 +34,11 @@ private[difftest] object DataMirror {
       val method = loadMethodOfObject("isVisible", "chisel3.reflect.DataMirror")
       method.exists(_.apply(data).asInstanceOf[Boolean])
     }
+
+    def tapAndRead(implicit si: SourceInfo): Data = {
```
poemonsense, line 38: should be polymorphic

src/main/scala/common/WiringControl.scala
```diff
@@ -61,15 +81,15 @@ object DifftestWiring {
     }
   }
 
-  def addSource[T <: Data](data: T, name: String): T = {
```
poemonsense, line 64: should be polymorphic

should not use `config`. Use a boolean as the third argument please. And add a function without third argument, default the third to false

klin02:
To simplify using other Methods of DifftestWiring, set isImplicit before addSource/addSink.

To make tapAndRead polymorphic, also add template to implicit class.

poemonsense:
why not add the isImplicit argument?

src/main/scala/util/DataMirror.scala
```diff
@@ -28,10 +29,16 @@ private[difftest] object DataMirror {
...
+  implicit class DataMirrorLoader[T <: Data](data: T) {
     def isVisible: Boolean = {
       val method = loadMethodOfObject("isVisible", "chisel3.reflect.DataMirror")
       method.exists(_.apply(data).asInstanceOf[Boolean])
     }
+
+    def tapAndRead(implicit si: SourceInfo): T = {
+      val method = loadMethodOfObject("tapAndRead", "chisel3.util.experimental.BoringUtils")
+      val argument: Seq[Any] = Seq(data, si)
+      method.get.apply(argument: _*).asInstanceOf[T]
```
poemonsense, line 41: Will this `get` cause errors in Chisel 3.6.x?

isVisible uses scala reflection because making isVisible false does not bring functional issues in Chisel 3.6.x, and we want to use the real isVisible in Chisel 6. Please understand the usage here before using it

src/main/scala/common/WiringControl.scala
```diff
@@ -61,13 +82,15 @@ object DifftestWiring {
...
+  def setConfig(isImplicit: Boolean) = WiringControl.isImplicit = isImplicit
+
   def addSource[T <: Data](data: T, name: String): T = {
     getWire(data, name).setSource()
     WiringControl.addSource(data, name)
     data
   }
 
-  def addSink[T <: Data](data: T, name: String): T = {
+  def addSink[T <: Data](data: T, name: String, config: GatewayConfig = GatewayConfig()): T = {
```
poemonsense, line 93: WiringControl is in difftest.common and should never depend on anything like GatewayConfig.

It should be the infrastructure. Its arguments should be simple.

src/main/scala/Gateway.scala
```diff
@@ -129,7 +131,13 @@ object Gateway {
 
   def collect(): GatewayResult = {
     val sink = if (config.needEndpoint) {
-      val endpoint = Module(new GatewayEndpoint(instances.toSeq, config))
+      val signals = instances.toSeq
+      val packed = WireInit(0.U.asTypeOf(MixedVec(signals.map(gen => UInt(gen.getWidth.W)))))
+      for ((data, idx) <- packed.zipWithIndex) {
+        DifftestWiring.addSink(data, s"gateway_$idx")
```
poemonsense, line 137: why not `DifftestWiring.addSink(data,  s"gateway_$idx", config.implicitWiring)` here? Then we don't need a config at DifftestWiring

poemonsense (commented):
I still do not understand why there has to be a config in DifftestWiring. It seems adding a third argument is enough

src/main/scala/Gateway.scala
klin02: If so we need to pass config.implicitWiring to addSource/addSink/createPendingWires.
We add setConfig because we can choose Wiring mode before use, it seems more confused to set isImplicit for every signal.

poemonsense: For a public API, it should be as simple as possible.

If you are going to let all DifftestWiring in Gateway (a user for the DifftestWiring API) use the same `isImplicit`, it is better to implement a wrapper in Gateway to call DifftestWiring.

klin02: Sure. I will have a try.

src/main/scala/common/WiringControl.scala
```diff
@@ -42,54 +45,76 @@ private class WiringInfo(val dataType: Data, val name: String) {
...
     if (isPending) {
-      Some((nSources == 0, dataType, name))
+      Some((nSources == 0, chiselTypeOf(data), name, isImplicit))
     } else {
       None
     }
   }
+
+  // (hasSource, data, name)
+  def toSourceTuple = (nSources != 0, data, name)
```
poemonsense, line 57: Why do we need this toSource tuple? `data` and `name` exists in the class definition and can be accessed directly

src/main/scala/common/WiringControl.scala
```diff
@@ -42,54 +45,76 @@ private class WiringInfo(val dataType: Data, val name: String) {
 
   def isPending: Boolean = nSources == 0 || nSinks == 0
   // (isSource, dataType, name)
-  def toPendingTuple: Option[(Boolean, Data, String)] = {
+  def toPendingTuple: Option[(Boolean, Data, String, Boolean)] = {
     if (isPending) {
-      Some((nSources == 0, dataType, name))
+      Some((nSources == 0, chiselTypeOf(data), name, isImplicit))
```
poemonsense, line 50: I doubt this chiselTypeOf. What will happen if an implicit data is pending and we call createPendingWires?

poemonsense (approved):

PR #420, opened 2024-07-16
top: add the exit output to exit the simulation

poemonsense:
This commit adds the exit output to the top-level IO of difftest. When it is a non-zero value, the simulation should exit. When all bits are ones, the simulation exits normally. Otherwise, it indicates there are some errors in the circuit.

### DIFF
--- a/src/test/vsrc/vcs/DifftestEndpoint.v
+++ b/src/test/vsrc/vcs/DifftestEndpoint.v
@@ -34,6 +34,7 @@ module DifftestEndpoint(
   input  wire [ 7:0] difftest_uart_out_ch,
   output wire        difftest_uart_in_valid,
   output wire [ 7:0] difftest_uart_in_ch,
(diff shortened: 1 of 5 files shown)


src/test/vsrc/vcs/DifftestEndpoint.v
```diff
@@ -192,11 +193,21 @@ always @(posedge clock) begin
   else begin
     n_cycles <= n_cycles + 64'h1;
 
-    // max cycles
-    if (max_cycles > 0 && n_cycles >= max_cycles) begin
-      $display("EXCEEDED MAX CYCLE: %d", max_cycles);
-      $finish();
-    end
+  // max cycles
```
klin02, line 196: Shall we keep the indentation?

poemonsense: did not notice it. I'll fix it

klin02 (approved):

PR #423, opened 2024-07-17
verification: concat assertion bits for wiring

poemonsense:
This commit changes single assertion bits to a concatenated UInt and uses this UInt for wiring. This change effectively reduces the number of wirings for large designs, probably improving the wiring performance.


poemonsense:
After this PR, XiangShan builds successfully with 4631 assertions.

klin02 (approved):

PR #424, opened 2024-07-17
gateway: add the config to exit on assertions

poemonsense:
This commit adds the exitOnAssertions config for Gateway with the command-line argument --difftest-config X.

When it is specified, difftest will automatically connect the assertion bit to the exit signal. The assertions are extracted by the extractor transform.


poemonsense:
This feature has been tested on NutShell and XiangShan. Do not know how we can add a CI test for this.

klin02 (approved):

PR #427, opened 2024-07-18
difftest: support dump and rebuild by profile

klin02:
In acclerated and multi-core verification cases, we need to adjust or build Difftest's Hardware side seperately.

We use json to record Difftest apply and finish Info of DUT. Then difftest can rebuild for single or multi cores according to Profile.

### DIFF
--- a/src/test/scala/DifftestMain.scala
+++ b/src/test/scala/DifftestMain.scala
@@ -19,6 +19,11 @@ package difftest
 import chisel3._
 import chisel3.stage._
 
+import java.nio.file.{Files, Paths}
+import scala.annotation.tailrec
+import org.json4s.DefaultFormats
+import org.json4s.native.JsonMethods.parse
+
 // Main class to generate difftest modules when design is not written in chisel.
 class DifftestTop extends Module {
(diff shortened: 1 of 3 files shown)


klin02:
I'm thinking is it appropriate to to generate SimTop in DifftestMain. If not, I can seperate it to a single file.

klin02:
Json profile will dump to $(NOOP_HOME)/build/generated-src/difftest_profile.json

Example usage of rebuilding Difftest：
make difftest_verilog PROFILE=../build/generated-src/difftest_profile.json NUM_CORES=2 CONFIG=ZSE

src/test/scala/DifftestMain.scala
```diff
@@ -54,6 +59,55 @@ class DifftestTop extends Module {
   DifftestModule.finish("demo")
 }
 
+// Generate simulation interface based on Profile describing the instantiated information of design
+class SimTop(ProfileName: String, numCores: Int) extends Module {
```
poemonsense, line 63: `profileName` instead of `ProfileName`

poemonsense (approved):
Others look good

poemonsense (approved):

PR #428, opened 2024-07-18
difftest: instantiate Delayer inside Endpiont

klin02:
With Hierarchical Wiring (require Chisel6 and difftestConfig H), we can collect signals from DUT to Difftest, like SimTop.xx.yy.

Previous we instantiate Delayer before addSource, the Delayer is instantiated inside DUT. This change move it after addSource, so Design and Verification code can be seperated when needEndpoint and Hierarchical Wiring.

When Hierarchical wiring, We use Endpoint by default.

### DIFF
--- a/src/main/scala/Batch.scala
+++ b/src/main/scala/Batch.scala
@@ -20,6 +20,7 @@ import chisel3.util._
 import difftest._
 import difftest.gateway.GatewayConfig
 import difftest.common.DifftestPerf
+import difftest.util.Delayer
 
 import scala.collection.mutable.ListBuffer
 
--- a/src/main/scala/Gateway.scala
+++ b/src/main/scala/Gateway.scala
@@ -19,6 +19,7 @@ import chisel3._
 import chisel3.util._
 import difftest._
 import difftest.common.DifftestWiring
+import difftest.util.Delayer
 import difftest.dpic.DPIC
 import difftest.squash.Squash
 import difftest.batch.{Batch, BatchIO}
@@ -50,7 +51,7 @@ case class GatewayConfig(
   def batchArgByteLen: (Int, Int) = if (isNonBlock) (3900, 92) else (7800, 192)
   def hasDeferredResult: Boolean = isNonBlock || hasInternalStep
   def needTraceInfo: Boolean = hasReplay
-  def needEndpoint: Boolean = hasGlobalEnable || hasDutZone || isBatch || isSquash
+  def needEndpoint: Boolean = hasGlobalEnable || hasDutZone || isBatch || isSquash || hierarchicalWiring
   def needPreprocess: Boolean = hasDutZone || isBatch || isSquash || needTraceInfo
   // Macros Generation for Cpp and Verilog
   def cppMacros: Seq[String] = {
@@ -100,7 +101,7 @@ case class GatewayResult(
 }
 
 object Gateway {
-  private val instances = ListBuffer.empty[DifftestBundle]
+  private val instanceWithDelay = ListBuffer.empty[(DifftestBundle, Int)]
   private var config = GatewayConfig()
 
   def setConfig(cfg: String): Unit = {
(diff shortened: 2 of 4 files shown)


src/main/scala/Gateway.scala
```diff
@@ -50,7 +50,7 @@ case class GatewayConfig(
   def batchArgByteLen: (Int, Int) = if (isNonBlock) (3900, 92) else (7800, 192)
   def hasDeferredResult: Boolean = isNonBlock || hasInternalStep
   def needTraceInfo: Boolean = hasReplay
-  def needEndpoint: Boolean = hasGlobalEnable || hasDutZone || isBatch || isSquash
+  def needEndpoint: Boolean = hasGlobalEnable || hasDutZone || isBatch || isSquash || hierarchicalWiring
```
poemonsense, line 53: Probably in the future we will by default use hierarchicalWiring as well as the endpoint module?

I'm simplt considering the future design of the difftest framework. We can think about it now.

src/main/scala/Gateway.scala
```diff
@@ -101,6 +101,7 @@ case class GatewayResult(
 
 object Gateway {
   private val instances = ListBuffer.empty[DifftestBundle]
+  private val delayList = ListBuffer.empty[Int]
```
poemonsense, line 104: It is a better practice to keep track of the bundle and delay in the same list (because they are always paired).

`ListBuffer.empty[(DifftestBundle, Int)]`

src/main/scala/common/Delay.scala
```diff
@@ -0,0 +1,61 @@
+/***************************************************************************************
+ * Copyright (c) 2020-2024 Institute of Computing Technology, Chinese Academy of Sciences
```
poemonsense, line 2: Add BOSC as well.

How about rename this file to `Delayer.scala`?

src/main/scala/common/Delay.scala
```diff
@@ -0,0 +1,61 @@
...
+ *          http://license.coscl.org.cn/MulanPSL2
+ *
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND,
+ * EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+ *
+ * See the Mulan PSL v2 for more details.
+ ***************************************************************************************/
+
+package difftest.common
```
poemonsense, line 16: What about package `difftest.util`?

difftest.common is for outside users to use the infrastructures provided by DiffTest. util is more likely for internal usage of DiffTest (probably I should move `object VerificationExtractor` to package common as well)

poemonsense (dismissed):

klin02:
Sure. I will make a change.

src/main/scala/Gateway.scala
klin02: When XiangShan generate verilog for FPGA ( make verilog ). It seems still need addSource/addSink, even there is nothing in DPIC module (wrapper by SYNTHESIS macro). 

Maybe we can set hierarchicalWiring when simulation ( add difftest-config H to Makefile), but still use fly-lines as default config.

src/main/scala/Batch.scala
```diff
@@ -20,6 +20,7 @@ import chisel3.util._
 import difftest._
 import difftest.gateway.GatewayConfig
 import difftest.common.DifftestPerf
+import difftest.util.Delayer
```
poemonsense, line 23: Unused import?

klin02: Used in https://github.com/OpenXiangShan/difftest/blob/c2419086d7ad0faca3570660d36cfd28ea3e1b9e/src/main/scala/Batch.scala#L199
https://github.com/OpenXiangShan/difftest/blob/c2419086d7ad0faca3570660d36cfd28ea3e1b9e/src/main/scala/Batch.scala#L300
and some other place.

poemonsense (approved):

PR #429, opened 2024-07-19
lightsss: parent manually kills child that not the oldest when wake up

Lemover:
Current Problem:
1. Non-oldest child process call exit(0) to kill itself. And the parent child doesn't waitpid it.
2. The non-oldest child gets stucked, and exit(0) is failed to over.
3. There 'whole' program can not stop correctly or there will be a zombie process.

Solution: 
The parent manually kills the non-oldest child processes in wake-up just like it does in do-fork.

### DIFF
--- a/src/test/csrc/common/lightsss.cpp
+++ b/src/test/csrc/common/lightsss.cpp
@@ -63,7 +63,7 @@ int LightSSS::do_fork() {
   // the original process
   else if (pid != 0) {
     slotCnt++;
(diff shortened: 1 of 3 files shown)


klin02:
Format check error. Please run 'make format' in difftest and push again ~

poemonsense (changes requested):
I'll test it manually and list results here

poemonsense (approved):

PR #432, opened 2024-08-02
Trace: support dump and load trace of difftest IOs

klin02:
Trace module will dump trace of IOs between DUT and difftest. With trace file and json profile, we can drive difftest without DUT, which will speed up repeated simulation without DUT modification, and support difftest modification and iteration.

Note we padding 100 trace at the end of traceFile when dumping, because when enable some features for acceleration, Difftest may finish comparision within some cycles after loading trace.

### DIFF
--- a/src/main/scala/Difftest.scala
+++ b/src/main/scala/Difftest.scala
@@ -118,6 +118,21 @@ sealed trait DifftestBundle extends Bundle with DifftestWithCoreid { this: Difft
     cpp.mkString("\n")
   }
 
+  def toTraceDeclaration: String = {
+    def byteWidth(data: Data) = (data.getWidth + 7) / 8 * 8
+    val cpp = ListBuffer.empty[String]
+    cpp += "typedef struct __attribute__((packed)) {"
+    elements.toSeq.reverse.foreach { case (name, data) =>
+      val (typeWidth, arrSuffix) = data match {
+        case v: Vec[_] => (byteWidth(v.head), s"[${v.length}]")
+        case _         => (byteWidth(data), "")
+      }
+      cpp += f"  ${s"uint${typeWidth}_t"}%-8s $name$arrSuffix;"
+    }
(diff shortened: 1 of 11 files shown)


src/main/scala/Difftest.scala
```diff
@@ -118,6 +118,21 @@ sealed trait DifftestBundle extends Bundle with DifftestWithCoreid { this: Difft
     cpp.mkString("\n")
   }
 
+  def toTraceDeclaration: String = {
+    def byteWidth(data: Data) = (data.getWidth + 7) / 8 * 8
+    val cpp = ListBuffer.empty[String]
+    cpp += "typedef struct __attribute__((packed)) {"
```
poemonsense, line 124: Maybe we should move these parsers to the util package in the future. The recursive bundle parser is used in several files, such as DPIC

src/test/csrc/difftest/difftrace.cpp
```diff
@@ -15,37 +16,42 @@ DiffTrace::DiffTrace(const char *_trace_name, bool is_read, uint64_t _buffer_siz
...
   }
-  memcpy(trace, buffer + buffer_count, sizeof(DiffTestState));
+  memcpy(trace, buffer + buffer_count, sizeof(T));
   buffer_count++;
   // printf("%lu...\n", buffer_count);
   return 0;
 }
 
-bool DiffTrace::trace_file_next() {
+template <typename T> bool DiffTrace<T>::trace_file_next() {
```
poemonsense, line 38: It's good to add the template.

There are some future ideas.

1) add an env wrapper. Currently this code parses NOOP_HOME at runtime. Previously we are adding the macro also at compile-time. The priority now is run-time argument, runtime env, and compile-time env.

2) add compression/decompression to the trace class. 

These are future work

poemonsense (dismissed):

klin02:
Trace will be dumped and loaded before Delayer to avoid delayed twice.

poemonsense (approved):

PR #434, opened 2024-08-07
pc: add compare for the first instr pc of a commit group.

NewPaulWalker:
Sometimes the dut or ref will wrongly jump to a pc, and if their reg state still match in the next instructions, this will make debugging very difficult. So I add the comparison to the pc.

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -318,6 +318,15 @@ int Difftest::step() {
     dut->event.valid = 0;
     dut->commit[0].valid = 0;
   } else {
+#if !defined(BASIC_DIFFTEST_ONLY) && !defined(CONFIG_DIFFTEST_SQUASH)
+    if (dut->commit[0].valid) {
+      dut_commit_first_pc = dut->commit[0].pc;
+      ref_commit_first_pc = proxy->pc;
+      if (dut_commit_first_pc != ref_commit_first_pc) {
+        pc_mismatch = true;
+      }
+    }
+#endif
     for (int i = 0; i < CONFIG_DIFF_COMMIT_WIDTH; i++) {
       if (dut->commit[i].valid) {
         if (do_instr_commit(i)) {
@@ -358,7 +367,7 @@ int Difftest::step() {
     return 1;
   }
 
-  if (proxy->compare(dut)) {
+  if (proxy->compare(dut) || pc_mismatch) {
 #ifdef FUZZING
     if (in_disambiguation_state()) {
       Info("Mismatch detected with a disambiguation state at pc = 0x%lx.\n", dut->trap.pc);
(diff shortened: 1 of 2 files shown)


klin02:
When define CONFIG_DIFFTEST_SQUASH, multi commits in a group will be squashed to a single one, which means next squashed commit[0].pc will mismatch previous REF pc.

Please add `ifndef CONFIG_DIFFTEST_SQUASH` when using pc_mismatch.

src/test/csrc/difftest/difftest.cpp
```diff
@@ -318,6 +318,17 @@ int Difftest::step() {
     dut->event.valid = 0;
     dut->commit[0].valid = 0;
   } else {
+#ifdef CPU_XIANGSHAN
```
klin02, line 321: Should not specify CPU_XIANGSHAN，this feature applys to other CPU, and this define can not tested by CI.

Makefile
```diff
@@ -234,7 +234,7 @@ ifeq ($(shell clang-format --version 2> /dev/null| cut -f3 -d' ' | tr '.' '_'),
 else
 	@echo "Required clang-format Version: $(CLANG_FORMAT_VER)"
 	@echo "Your Version: $(shell clang-format --version)"
-	@echo "Please run \"pip install --user clang-format==$(CLANG_FORMAT_VER)", then set PATH manually"
+	@echo "Please run \"pip install --user clang-format==$(CLANG_FORMAT_VER)\", then set PATH manually"
```
klin02, line 237: I move this change to PR #435, which should be seperated from feature change.

poemonsense:
First pc of REF is copied from the DUT.

src/test/csrc/difftest/difftest.cpp
NewPaulWalker: > Should not specify CPU_XIANGSHAN，this feature applys to other CPU, and this define can not tested by CI.

Ok, I'll change it later. Previously I thought that the Nutshell behavior didn't support this feature. It now appears that the ci failed because turn onCONFIG_DIFFTEST_SQUASH?

NewPaulWalker:
> First pc of REF is copied from the DUT.

Yes, when initializing difftest, the REF state is copied from the DUT. What this pr wants to do is to compare whether the first pc in every commit group matches.

src/test/csrc/difftest/difftest.cpp
klin02: Yes.

poemonsense:
> > First pc of REF is copied from the DUT.
> 
> 
> 
> Yes, when initializing difftest, the REF state is copied from the DUT. What this pr wants to do is to compare whether the first pc in every commit group matches.

I see. So the problem is that difftest is missing the comparison of pc. Right?

I remember there was the comparison. Maybe it is removed at some point.

NewPaulWalker:
> I see. So the problem is that difftest is missing the comparison of pc. Right?

Yes.

klin02 (approved):

PR #438, opened 2024-08-12
Difftest: check timeout for both commit and step

klin02:
Since difftest step function may not be called every cycle, we use cycleCnt recorded in trapEvent instead of ticks, to check timeout for Instr commits.

However, commit timeout will only be checked when difftest checks, which is triggered by difftest_step. We also check if step is 0 for more than stuck_limit for both vcs and emu.

### DIFF
--- a/src/test/csrc/difftest/difftest.h
+++ b/src/test/csrc/difftest/difftest.h
@@ -280,28 +280,31 @@ protected:
   DiffTrace<DiffTestState> *difftrace = nullptr;
 
 #ifdef CONFIG_DIFFTEST_BATCH
-  const uint64_t commit_storage = CONFIG_DIFFTEST_BATCH_SIZE;
+  static const uint64_t commit_storage = CONFIG_DIFFTEST_BATCH_SIZE;
 #else
-  const uint64_t commit_storage = 1;
+  static const uint64_t commit_storage = 1;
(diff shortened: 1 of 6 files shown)


klin02:
This change need difftest_step always passed to tb_top, even when simv_step() is trigger internally. That is done by PR #437

klin02:
Commit timeout cannot simply checked by step(always 1 with Default Config), and needs Difftest semantic infos, such as WFI, so we check it inside Difftest.
However, sometimes for acceleration, difftest_step function will only called when needed, so we also need to check if cpu is stucked and will never calls Difftest again. We support step stuck checked in both emu and vcs.

src/test/csrc/difftest/difftest.h
```diff
@@ -279,29 +292,11 @@ class Difftest {
 protected:
   DiffTrace<DiffTestState> *difftrace = nullptr;
 
-#ifdef CONFIG_DIFFTEST_BATCH
-  const uint64_t commit_storage = CONFIG_DIFFTEST_BATCH_SIZE;
-#else
-  const uint64_t commit_storage = 1;
-#endif // CONFIG_DIFFTEST_BATCH
```
klin02, line 286: Since cycleCnt is recoreded for every group (a group means DiffState in same cycle) in Batch. We remove this.

poemonsense (approved):

PR #439, opened 2024-08-13
iotrace: add iotrace dump and load zstd compress support

xiaokamikami:
Adding the difftest trace feature enables ZSTD for streaming compression, which compresses the trace volume by about 98%, enabling printing more cycle traces to disk

### DIFF
--- a/src/test/csrc/difftest/difftrace.cpp
+++ b/src/test/csrc/difftest/difftrace.cpp
@@ -2,10 +2,6 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
-#ifdef CONFIG_IOTRACE_ZSTD
-static DiffTraceZstd *trace_zstd = new DiffTraceZstd;
-#endif // CONFIG_IOTRACE_ZSTD
-
 template <typename T>
 DiffTrace<T>::DiffTrace(const char *_trace_name, bool is_read, uint64_t _buffer_size) : is_read(is_read) {
   buffer_size = _buffer_size;
@@ -124,19 +120,19 @@ template class DiffTrace<DiffTestState>;
 #ifdef CONFIG_IOTRACE_ZSTD
 void DiffTraceZstd::diff_zstd_next(const char *file_name, bool is_read) {
   if (is_read) {
-    if (io_trace_file_i.is_open()) {
-      io_trace_file_i.close();
+    if (io_trace_file.is_open()) {
+      io_trace_file.close();
     }
-    io_trace_file_i.open(file_name, std::ios::binary);
-    if (io_trace_file_i.is_open() == false) {
+    io_trace_file.open(file_name, std::ios::binary);
+    if (io_trace_file.is_open() == false) {
       printf("No more trace files.End simulation\n");
       exit(0);
     }
   } else {
-    if (io_trace_file_o.is_open()) {
-      io_trace_file_o.close();
+    if (io_trace_file.is_open()) {
+      io_trace_file.close();
     }
-    io_trace_file_o.open(file_name, std::ios::binary);
+    io_trace_file.open(file_name, std::ios::binary);
   }
 }
 
@@ -153,7 +149,7 @@ void DiffTraceZstd::diff_IOtrace_dump(const char *str, uint64_t len) {
     return;
   }
 
-  io_trace_file_o.write(outputBuffer.data(), compressedSize);
+  io_trace_file.write(outputBuffer.data(), compressedSize);
   ZSTD_freeCCtx(trace_cctx);
 }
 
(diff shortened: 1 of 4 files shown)


klin02:
Also add CI check for ZSTD feature.

src/test/csrc/difftest/difftrace.cpp
```diff
@@ -2,10 +2,14 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+#ifdef CONFIG_IOTRACE_ZSTD
+static DiffTraceZstd *trace_zstd = new DiffTraceZstd;
```
klin02, line 6: Seems a little strange to call a static external instance from template class. 
Maybe declare this valuable inside Difftrace?

src/test/csrc/difftest/difftrace.cpp
```diff
@@ -49,10 +53,36 @@ template <typename T> bool DiffTrace<T>::trace_file_next() {
...
-  snprintf(filename, 128, "%s/%lu.%s", dirname, trace_index, prefix);
+#else
+  const char *prefix = "zstd";
+#endif // CONFIG_IOTRACE_ZSTD
+  snprintf(file_name, 128, "%s/%lu.%s", dirname, trace_index, prefix);
+  trace_index++;
+}
+
+template <typename T> bool DiffTrace<T>::trace_file_next() {
+  char *filename = (char *)malloc(128);
```
klin02, line 66: Maybe it's better to let next_file_name return a char ptr directly.
Rather than malloc outside and modify it in another function, because `128` in snprintf corresponds to here.

src/test/csrc/difftest/difftrace.cpp
```diff
@@ -71,19 +101,135 @@ template <typename T> bool DiffTrace<T>::trace_file_next() {
...
+    io_trace_file_i.open(file_name, std::ios::binary);
+    if (io_trace_file_i.is_open() == false) {
+      printf("No more trace files.End simulation\n");
+      exit(0);
+    }
+  } else {
+    if (io_trace_file_o.is_open()) {
+      io_trace_file_o.close();
+    }
+    io_trace_file_o.open(file_name, std::ios::binary);
```
klin02, line 139: Maybe io_trace_file is OK? We will not read and write at the same time, seems no need to declare _o and _i both.

src/test/csrc/difftest/difftrace.cpp
```diff
@@ -71,19 +101,135 @@ template <typename T> bool DiffTrace<T>::trace_file_next() {
...
+    std::cout << "there might be some data left within internal buffers" << std::endl;
+  }
+  if (input.pos == input.size) {
+    std::cout << "once a block has been processed, more data is needed" << std::endl;
+  }
+  if (ZSTD_isError(ret)) {
+    std::cerr << "Decompression error: " << ZSTD_getErrorName(ret) << std::endl;
+    exit(0);
+  }
+#endif // IOTRACE_ZSTD_DEBUG
```
klin02, line 226: If test OK, remember to remove these printf.

src/test/csrc/difftest/difftrace.h
```diff
@@ -5,13 +5,20 @@
...
+#endif // CONFIG_IOTRACE_ZSTD
 
+const uint64_t trace_buffer_size = 1024 * 1024;
 template <typename T> class DiffTrace {
 public:
   char trace_name[32];
   bool is_read;
 
-  DiffTrace(const char *trace_name, bool is_read, uint64_t buffer_size = 1024 * 1024);
+  DiffTrace(const char *trace_name, bool is_read, uint64_t buffer_size = trace_buffer_size);
```
klin02, line 21: Keep it configurable. If trace_buffer_size is needed by class DiffTraceZSTD, you can create instance inside this function.

src/test/csrc/difftest/difftrace.cpp
xiaokamikami: The behavior of returning char ptr should be unsafe

src/test/csrc/difftest/difftrace.cpp
```diff
@@ -71,19 +97,123 @@ template <typename T> bool DiffTrace<T>::trace_file_next() {
...
+    io_trace_file.open(file_name, std::ios::binary);
+    if (io_trace_file.is_open() == false) {
+      printf("No more trace files.End simulation\n");
+      exit(0);
+    }
+  } else {
+    if (io_trace_file.is_open()) {
+      io_trace_file.close();
+    }
+    io_trace_file.open(file_name, std::ios::binary);
```
klin02, line 135: Seems replicated code in if and else.

src/test/csrc/difftest/difftrace.cpp
```diff
@@ -71,19 +97,123 @@ template <typename T> bool DiffTrace<T>::trace_file_next() {
     uint64_t read_bytes = fread(buffer, sizeof(T), buffer_size, file);
     assert(read_bytes == buffer_size);
     fclose(file);
-    buffer_count = 0;
+#else
+    buffer = (T *)calloc(buffer_size, sizeof(T));
```
klin02, line 101: Remember to free `malloc` when this function called repeatedly. Also check for other `malloc` and `calloc`.

klin02: Also keep Loading Info here, which notifies us Trace is working.

xiaokamikami: OK

klin02 (approved):

PR #440, opened 2024-08-15
Validate: recover needUpdate for TrapEvent

klin02:
TrapEvent may not raise hasTrap when any other bundle valid, so we recover needUpdate to mark there is valid TrapEvent at this cycle.

In Validate, we use getValid to mark valid for single bundle, and needUpdate to mark valid for this cycle. We will transmit trapEvent when hasTrap, or other bundles valid.

### DIFF
--- a/src/main/scala/Bundles.scala
+++ b/src/main/scala/Bundles.scala
@@ -95,6 +95,8 @@ class TrapEvent extends DifftestBaseBundle {
 
   val code = UInt(32.W)
   val pc = UInt(64.W)
+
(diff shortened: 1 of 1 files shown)


poemonsense:
trap event also contains cycleCnt and instrCnt information. We need to fix them before changing the transmission of trap event. Do you check them?

klin02:
Sure. When hasTrap, we transmit TrapEvent to finish simulation, when other bundle valid, we transmit TrapEvent to provide excactly cycleCnt and instrCnt

poemonsense (approved):

PR #441, opened 2024-08-15
Difftest: support compiled with Chisel6 and Chisel3

klin02:


klin02:
To generate seperated sv files according to JsonProfile, we use Chisel6 to compile Difftest by Default.

As some features, such as DataMirror, only supports Chisel6, now we partly make it compatible for Chisel3 using reflection dynamically. However, it is still hard to use some features accomplished in Chisel6.

Since XiangShan and NutShell have already purged Chisel3, does Difftest still need to keep Chisel3 compatible?

poemonsense (commented):
We need to maintain basic support for Chisel 3. For example, the current TopMain should compile correctly with Chisel 3.

Advanced top may only compile with Chisel 6.

Probably we can pass crossValue from command line and let use MFC=1 by default. Different tests can use different options.

klin02:
Sure. I will make a change

poemonsense (approved):

PR #444, opened 2024-08-19
l2tlb: add pbmt attribute checker for Svpbmt

Maxpicca-Li:


poemonsense (dismissed):

poemonsense:
Will Svpbmt affect the TLB translation results? Are they required for diff?

Maxpicca-Li:
> Will Svpbmt affect the TLB translation results? Are they required for diff?

Yes, because pbmt is a part of PTE, like `perm`.

poemonsense (approved):

PR #447, opened 2024-08-27
misc: add argument --dramsim3-ini to set dramsim3 config file

tastynoob:
add argument --dramsim3-ini to pass the dramsim3 config file, no need to modify makefile and recompile

### DIFF
--- a/src/test/csrc/common/ram.cpp
+++ b/src/test/csrc/common/ram.cpp
@@ -289,10 +289,6 @@ MmapMemory::MmapMemory(const char *image, uint64_t n_bytes) : SimMemory(n_bytes)
     img_size = reader->read_all(ram, memory_size);
     delete reader;
   }
-
-#ifdef WITH_DRAMSIM3
-  dramsim3_init();
-#endif
 }
 
 MmapMemory::~MmapMemory() {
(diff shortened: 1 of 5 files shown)


poemonsense (changes requested):
1) does not make sense to change MmapMemory

2) ~~does not see how this argument affects the config of dramsim3~~ already fixed now

tastynoob:
> 1. does not make sense to change MmapMemory
> 2. ~does not see how this argument affects the config of dramsim3~ already fixed now

MmapMemory doesn't use dramsim3?

poemonsense:
> > 1. does not make sense to change MmapMemory
> > 2. ~does not see how this argument affects the config of dramsim3~ already fixed now
> 
> MmapMemory doesn't use dramsim3?

I see why you changed MmapMemory now. It is because the `dramsim3_init` was previously called in the constructor of MmapMemory, which is totally wrong. I think we should move this call outside the constructor and call it after `new MmapMemory`.

In emu.cpp:
```c++
    } else {
      init_ram(args.image, ram_size);
#ifdef WITH_DRAMSIM3
      dramsim3_init(args.dramsim3_ini);
#endif
    }
```

Similarly in src/test/csrc/vcs/vcs_main.cpp, we also calls dramsim3_init after init_ram.

Then the code looks better.

poemonsense:
Actually, dramsim3 should not be in ram.cpp now because it is only a behavior model for DRAM. ram.cpp handles memory read/write data but dramsim3 handles only the DRAM latency.

I'll re-organize the code after (or before?) you add the ini argument.

poemonsense (commented):
Thanks for the work. This looks good now.

However, VCS simulation may not work after this change. I think it also requires a call to dramsim_init after init_ram?

poemonsense:
We can use the default nullptr for config in VCS. Then we don't need to pass the command line argument

tastynoob:
> 我们可以在 VCS 中使用默认的 nullptr 进行配置。这样我们就不需要传递命令行参数了

If vcs simulate no need to test performance, this make sense

poemonsense:
> > 我们可以在 VCS 中使用默认的 nullptr 进行配置。这样我们就不需要传递命令行参数了
> 
> If vcs simulate no need to test performance, this make sense

Yes, but we still need to call `dramsim3_init(nullptr)` in vcs_main.cpp when `WITH_DRAMSIM3` is defined. Otherwise, VCS cannot build with dramsim3.

Please help add it and then we can merge it

poemonsense (dismissed):
Look good to me now.

The CI may fail due to the Verilator version bump (see https://github.com/OpenXiangShan/difftest/pull/449). If CI fails, please rebase this branch after that PR merges

poemonsense (approved):

PR #448, opened 2024-08-27
difftest: set valid to false when using dontCare

poemonsense:


poemonsense:
https://github.com/OpenXiangShan/xs-env/pull/49 bumps the default verilator version to v5.028, which includes a time checker that aborts the simulation if we start the simulation with a nonzero time. This is caused by the sc_time_stamp function in the source code.

This change aborts the CI as well. I'm going to remove it in another PR

PR #452, opened 2024-08-28
fix(l1tlb_check): correct the check for only VS stage

cebarobot:
As a legacy issue of supporting Sv48, the default level of do_s2xlate when mode=0 is incorrect. Also, for only VS stage case (onlyS1), it should not do_s2xlate at all. This patch fixes these problems.


good-circle (approved):

poemonsense:
Is this OK for merging? I'll squash and merge it

PR #454, opened 2024-09-02
feat(Smrnmi): add support to raise non-maskable interrupt

lewislzh:


src/test/csrc/difftest/refproxy.h
```diff
@@ -93,6 +93,11 @@ class RefProxyConfig {
   bool enable_store_log = false;
 };
 
+#ifdef CPU_XIANGSHAN
```
poemonsense, line 96: The `raise_nmi_intr` should be declared as an optional interface. See `#define REF_OPTIONAL(f)  ...`

poemonsense (commented):
Will NMI interrupt and normal interrupt occur at the same time?

Generally speaking, NMI should be implemented as an optional feature in DiffTest. For REFs with nmi intr, they should implement the raise_nmi_intr interface. Otherwise, if they do not implement it, NMI will be skipped.

Therefore, probably you should refer to the implementation of guided_exec (https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/difftest/refproxy.h#L228-L230) and follow similar ideas to implement `Difftest::do_interrupt`. For example, check if the raise_nmi_intr exists, and then decide whether to call it. Using `CPU_XIANGSHAN` to check the existence of nmi_intr is a bad idea.

src/test/csrc/difftest/difftest.cpp
```diff
@@ -398,6 +398,9 @@ int Difftest::step() {
 
 void Difftest::do_interrupt() {
   state->record_interrupt(dut->event.exceptionPC, dut->event.exceptionInst, dut->event.interrupt);
+#ifdef CPU_XIANGSHAN
+  proxy->raise_nmi_intr(dut->event.hasNMI);
```
poemonsense, line 402: Something like:

```c++
if (proxy->raise_nmi_intr) {
  // call it
}
else {
  // raise some warnings and exit or simply leave it empty
}
```

lewislzh: I have changed nmi as an optional feature like guided_exec.

lewislzh: NMI interrupt and normal interrupt can occur at the same time, and NMI interrupt has the highest priority.

src/test/csrc/difftest/refproxy.h
```diff
@@ -225,6 +226,12 @@ class RefProxy : public AbstractRefProxy {
     }
   }
 
+  inline void trigger_nmi(bool hasNMI) {
+    if (raise_nmi_intr) {
+      raise_nmi_intr(hasNMI);
+    }
```
poemonsense, line 232: Adding an else and saying raise_nmi_intr has not been supported by the REF?

poemonsense (approved):
Look good to me.

However, it may be better to add the else for debugging usages. Otherwise, if some REF does not support NMI, it would be difficult for the user to know that it is probably due to the missing definition in the REF.

PR #460, opened 2024-09-05
difftest: automatically set the overwrite-size of new gpct

xiaokamikami:
New versions of gcpt allow you to set oversize automatically. if you don't specify size manually, gcpt will look for an oversize flag，on vcs/palldium  using --set_overwrite_autoset enable auto set oversize，on emu using --overwrite-auto

### DIFF
--- a/src/test/csrc/vcs/vcs_main.cpp
+++ b/src/test/csrc/vcs/vcs_main.cpp
@@ -44,7 +44,6 @@ static bool enable_difftest = true;
 static uint64_t max_instrs = 0;
 static char *workload_list = NULL;
 static uint32_t overwrite_nbytes = 0xe00;
-static bool overwrite_auto = false;
 struct core_end_info_t {
   bool core_trap[NUM_CORES];
   double core_cpi[NUM_CORES];
@@ -81,7 +80,6 @@ extern "C" void set_overwrite_autoset() {
   }
   // Get the lower four bytes
   fseek(fp, 4, SEEK_SET);
-  uint32_t data = 0;
   fread(&overwrite_nbytes, sizeof(uint32_t), 1, fp);
   fclose(fp);
 }
--- a/src/test/csrc/vcs/vcs_main.cpp
+++ b/src/test/csrc/vcs/vcs_main.cpp
@@ -43,7 +43,8 @@ static char *gcpt_restore_bin = NULL;
 static bool enable_difftest = true;
 static uint64_t max_instrs = 0;
 static char *workload_list = NULL;
-static uint64_t overwrite_nbytes = 0xe00;
+static uint32_t overwrite_nbytes = 0xe00;
+static bool overwrite_auto = true;
 struct core_end_info_t {
   bool core_trap[NUM_CORES];
   double core_cpi[NUM_CORES];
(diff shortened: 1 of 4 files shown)


src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -70,11 +71,26 @@ extern "C" void set_flash_bin(char *s) {
...
+    fseek(fp, 4, SEEK_SET);
+    if (fread(&overwrite_nbytes, sizeof(uint32_t), 1, fp) != 1) {
+      fclose(fp);
+      printf("auto read gcpt overwrite size fiald, need manual set_overwrite_nbytes\n");
+    }
+    fclose(fp);
+    if (overwrite_nbytes > 1024 * 1024) {
+      printf("the workload you are using may not support overwrite automatically\n");
+    }
+    printf("gcpt auto overwrite size :%x\n", overwrite_nbytes);
```
poemonsense, line 92: Format the printf please. "size: " instead of "size :"

poemonsense (approved):
I think mixing in the dirty code for checkpoints is a bad idea. I suggest rethinking how the support for checkpoints should be implemented.

However, I'm ok to approve it if you can rearrange the multiple printf's. I don't understand why you need so many printf's

poemonsense (changes requested):
Why is the PR titled coverage?

xiaokamikami:
> I think mixing in the dirty code for checkpoints is a bad idea. I suggest rethinking how the support for checkpoints should be implemented.
> 
> However, I'm ok to approve it if you can rearrange the multiple printf's. I don't understand why you need so many printf's



> I think mixing in the dirty code for checkpoints is a bad idea. I suggest rethinking how the support for checkpoints should be implemented.
> 
> However, I'm ok to approve it if you can rearrange the multiple printf's. I don't understand why you need so many printf's

I have the code to print gcpt siz elsewhere, so I'm going to remove the printf here

xiaokamikami:
> Look good to me now.
> 
> Only VCS supports this? Do we need to add support for verilator?

I've also added support in case someone needs to run a new checkpoint on emu

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -43,7 +43,8 @@ static char *gcpt_restore_bin = NULL;
 static bool enable_difftest = true;
 static uint64_t max_instrs = 0;
 static char *workload_list = NULL;
-static uint64_t overwrite_nbytes = 0xe00;
+static uint32_t overwrite_nbytes = 0xe00;
+static bool overwrite_auto = false;
```
poemonsense, line 47: This bool seems useless?

src/test/csrc/verilator/emu.cpp
```diff
@@ -407,6 +410,13 @@ Emulator::Emulator(int argc, const char *argv[])
   }
 
   if (args.gcpt_restore) {
+        if (args.overwrite_nbytes_autoset) {
+            FILE *fp = fopen(gcpt_restore_bin, "rb");
+            fseek(fp, 4, SEEK_SET);
+            fread(&args.overwrite_nbytes, sizeof(uint32_t), 1, fp);
+            fclose(fp);
+
```
poemonsense, line 418: Should indent using 2 spaces instead of 4 spaces. Also delete this empty line please

poemonsense (dismissed):

poemonsense (approved):
Thank you for the great work

PR #464, opened 2024-09-12
feat(ram): increase memory size to 8TB

good-circle:


klin02 (approved):

config/config.h
```diff
@@ -36,7 +36,11 @@
 // -----------------------------------------------------------------------
 
 // emulated memory size (Byte)
+#if defined(CPU_NUTSHELL) || defined(CPU_ROCKET_CHIP)
 #define DEFAULT_EMU_RAM_SIZE (8 * 1024 * 1024 * 1024UL) // 8 GB
+#elif defined(CPU_XIANGSHAN)
```
poemonsense, line 41: Should be like this: 

```c++
#if defined(CPU_XIANGSHAN)
// 8TB
#else
// default 8GB
#endif
```

Otherwise, other CPUs cannot run with difftest.

PR #465, opened 2024-09-13
fix(csr): add support non-register interrupts pending

sinceforYy:
This PR adds support for non-register interrupts pending to copy the interrupt to NEMU. 
The interrupts come from Count overflow and Platform-Level Interrupt Controller such as PLIC, CLINT, IMSIC. 
Thus, we could diff xip csr registers of XiangShan and NEMU.

### DIFF
--- a/src/main/scala/Bundles.scala
+++ b/src/main/scala/Bundles.scala
@@ -301,6 +301,17 @@ class RunaheadRedirectEvent extends DifftestBaseBundle with HasValid {
   val checkpoint_id = UInt(64.W)
 }
 
+class NonRegInterruptPendingEvent extends DifftestBaseBundle with HasValid {
+  val platformIRPMeip = Bool()
+  val platformIRPMtip = Bool()
+  val platformIRPMsip = Bool()
+  val platformIRPSeip = Bool()
+  val platformIRPStip = Bool()
+  val platformIRPVseip = Bool()
+  val platformIRPVstip = Bool()
+  val localCounterOverflowInterruptReq = Bool()
+}
+
 class TraceInfo extends DifftestBaseBundle with HasValid {
   val in_replay = Bool()
   val trace_head = UInt(16.W)
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -316,6 +316,10 @@ int Difftest::step() {
   }
 #endif
 
+#ifdef CONFIG_DIFFTEST_NONREGINTERRUPTPENDINGEVENT
+  do_non_reg_interrupt_pending();
+#endif
+
   num_commit = 0; // reset num_commit this cycle to 0
   if (dut->event.valid) {
     // interrupt has a higher priority than exception
@@ -361,11 +365,6 @@ int Difftest::step() {
 
   if (num_commit > 0) {
     state->record_group(dut->commit[0].pc, num_commit);
-  }
-
-  // FIXME: the following code is dirty
-  if (dut->csr.mip != proxy->csr.mip) { // Ignore difftest for MIP
-    proxy->csr.mip = dut->csr.mip;
   }
 
   if (apply_delayed_writeback()) {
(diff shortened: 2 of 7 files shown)


src/test/csrc/difftest/refproxy.h
```diff
@@ -342,6 +347,17 @@ struct ExecutionGuide {
   uint64_t jump_target;
 };
 
+struct NoRegInterruptPending {
```
poemonsense, line 350: This struct is the same as the generated class in diffstate.h. May reuse the generated one

src/test/csrc/difftest/difftest.cpp
```diff
@@ -316,6 +316,8 @@ int Difftest::step() {
   }
 #endif
 
+  do_no_reg_interrupt_pending();
```
poemonsense, line 319: We need to handle the case when this feature is not supported. Thus, please use the macro to wrap the code. Please refer to other diff checkers (LR/SC, store, ...) and use ifdef.

src/main/scala/Bundles.scala
```diff
@@ -301,6 +301,17 @@ class RunaheadRedirectEvent extends DifftestBaseBundle with HasValid {
   val checkpoint_id = UInt(64.W)
 }
 
+class NoRegInterruptPendingEvent extends DifftestBaseBundle {
```
poemonsense, line 304: Please add this class to the README of DiffTest and explain its usage.

For the missing bundles added earlier, add them to README with explanations as well

poemonsense (commented):
We need explanations for these bundles in README and detailed introduction in this PR. This helps others including the reviewers to understand them

sinceforYy:
@poemonsense For failed CI, should I add supoort in NutShell and Rocket-chip?

poemonsense:
> @poemonsense For failed CI, should I add supoort in NutShell and Rocket-chip?

There's no need. You should wrap the code under macros. Then there will not be any errors.

src/test/csrc/difftest/refproxy.h
```diff
@@ -145,7 +152,8 @@ class RefProxyConfig {
   f(ref_guided_exec, difftest_guided_exec, void, void*)                       \
   f(raise_nmi_intr, difftest_raise_nmi_intr, void, bool)                      \
   f(ref_virtual_interrupt_is_hvictl_inject, difftest_virtual_interrupt_is_hvictl_inject, void, bool)        \
-  f(disambiguation_state, difftest_disambiguation_state, int, )
+  f(disambiguation_state, difftest_disambiguation_state, int, )               \
+  REF_NON_REG_INTERRUPT_PENDING(f)
```
poemonsense, line 156: It is already defined in the REF_OPTIONAL. There's no need to wrap it under the macro. DiffTest will handle the case when this function is not defined by the REF

src/test/csrc/difftest/refproxy.h
```diff
@@ -243,6 +251,12 @@ class RefProxy : public AbstractRefProxy {
     }
   }
 
+#ifdef CONFIG_DIFFTEST_NONREGINTERRUPTPENDINGEVENT
+  inline void non_reg_interrupt_pending(DifftestNonRegInterruptPendingEvent &nonRegInterruptPending) {
+    ref_non_reg_interrupt_pending(&nonRegInterruptPending);
```
poemonsense, line 256: Probably the better way is not to wrap this in macro because this code is for ref but the macro is controlled by dut. This file is expected to provide the abstract interface for any DUT to contact REF.

Maybe this is better (no need for else because REFs are allowed not to implement it, in which case pending interrupts may cause diff in MIPs, and the user should implement this interface after they debug it):

```c
if (ref_non_reg_interrupt_pending) {
  ref_non_reg_interrupt_pending(noRegInterruptPending);
}
```

poemonsense: I'm also considering that, should we always sync it to REF at every difftest tick? This may cause unnecessary function calls because if their values do not change, it seems there will not be any differences between the DUT and REF. 

Or, is it true that we only need to sync the values to REF if any one bit of them are true? If they are always zeros, REFs' bits will also be zeros.

These policies can be implemented by checking them in C++. But the better way seems to be adding a valid bit for the bundle. Only when the valid is asserted (the hardware detected any necessary changes to them), the software syncs it to the REF.

poemonsense (commented):
Some comments for improving the code quality.

For the sync policy, we may discuss more and you may discuss them with your group as well. I may miss the target conditions so please feel free to correct me.

src/test/csrc/difftest/refproxy.h
```diff
@@ -342,6 +349,17 @@ struct ExecutionGuide {
   uint64_t jump_target;
 };
 
+struct NonRegInterruptPending {
```
poemonsense, line 352: Why is the tip the first element instead of the eip as in Bundle.scala?

If their orders are the same, we can use `&(dut->non_reg_interrupt_pending.platformIRPMtip)` to get the pointer for REF. Then we don't need to define the struct here.

poemonsense (commented):
Thank you for the work. Look good to me except the struct definition.

If you think this struct is required, I'm open to merge it.

src/test/csrc/difftest/refproxy.h
sinceforYy: They are in order.
But there is a problem here. If struct is not defined, I can't get dut->non_reg_interrupt_pending in non_reg_interrupt_pending().

poemonsense: I see. Make sense.

Using a `void *` works but it will lose the struct information.

poemonsense (approved):

src/test/csrc/difftest/refproxy.h
sinceforYy: Yes. They should be in the same order, but currently they are not. This is my fault, the new commit was not force-pushed. I will open a new PR to fix it.

poemonsense: Thank you for the patience.

PR #467, opened 2024-09-13
Difftest: Use file API compatible with Java8

forever043:


poemonsense (commented):
I don't think we should support JDK8

forever043:
> I don't think we should support JDK8

most EDA tools runs in CentOS 7, and the default JRE was 1.8.0

If the cost is not too high, we should try not to rely too much on the Java version

poemonsense:
> > I don't think we should support JDK8
> 
> most EDA tools runs in CentOS 7, and the default JRE was 1.8.0
> 
> If the cost is not too high, we should try not to rely too much on the Java version

I think some submodules in XiangShan support only JDK>=11 as well. Are they fixed?

How can we test it? Probably we need to add a test for the environment if we are going to support them.

forever043:
> I think some submodules in XiangShan support only JDK>=11 as well. Are they fixed?
 
for my testcase, the XiangShan jar (generated by `make test-jar` or `make jar`) can generate verilog successfullly with the default jre1.8.0 in CentOS7.

> How can we test it? Probably we need to add a test for the environment if we are going to support them.

that is a good question. I'm concerned that not all of our CI jobs can run in an old OS distribution.

maybe we can add a simple `compile task` running in old OS, in which we can do the java compile/assembly work like `make jar/test-jar` in XiangShan. This will ensure our code do not use these **newer** APIs.

klin02:
Maybe we can install JDK11 on Centos7? I think it will be better to keep same JDK version between different machines. And it seems JDK11 faster than JDK8?

poemonsense:
> maybe we can add a simple `compile task` running in old OS, in which we can do the java compile/assembly work like `make jar/test-jar` in XiangShan. This will ensure our code do not use these **newer** APIs.

Yes. Compiling is enough. https://github.com/OpenXiangShan/difftest/blob/master/.github/workflows/main.yml#L22-L25. For example, ` mill -i design[chisel].compile`

poemonsense:
> Maybe we can install JDK11 on Centos7? I think it will be better to keep same JDK version between different machines. And it seems JDK11 faster than JDK8?

I agree. Installing JDK is quite easy. Simply download a tar.gz and unzip it. For example, https://www.graalvm.org/latest/docs/getting-started/linux/

klin02:
I think for this API, it's OK to make it compatible for more JDK version without extra cost. 
But when API conflicts with older version, there is no need to make it compatible anymore.

poemonsense (approved):
I'm open for merging.

forever043:
> ... But when API conflicts with older version, there is no need to make it compatible anymore.
totally agree, if we need to much effort to make the old version compatible, it is time to drop it :)
for now, it's still ok (at least for my scenario)

PR #469, opened 2024-09-14
fix(csr): fix struct non-reg interrupt pending to order

sinceforYy:
Adjust the order.


poemonsense (approved):

poemonsense:
We will consider how to better sync the data structure between difftest and REF in the future.

sinceforYy:
> We will consider how to better sync the data structure between difftest and REF in the future.

I'd be happy to help if needed.

PR #473, opened 2024-09-24
DPIC: change Batch param type to support FPGA/PCIe

xiaokamikami:
A struct is generated for the batch dpic function when generating the batch code，and the variable type of the batch entry is changed when the fpga is enabled

### DIFF
--- a/src/main/scala/DPIC.scala
+++ b/src/main/scala/DPIC.scala
@@ -31,18 +31,22 @@ abstract class DPICBase(config: GatewayConfig) extends ExtModule with HasExtModu
   val enable = IO(Input(Bool()))
   val dut_zone = Option.when(config.hasDutZone)(IO(Input(UInt(config.dutZoneWidth.W))))
   val step = Option.when(config.hasInternalStep)(IO(Input(UInt(config.stepWidth.W))))
+  val fpga = config.fpgaEnable
 
   def getDirectionString(data: Data): String = {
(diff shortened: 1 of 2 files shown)


src/main/scala/DPIC.scala
```diff
@@ -31,18 +31,22 @@ abstract class DPICBase(config: GatewayConfig) extends ExtModule with HasExtModu
...
-  def getDPICArgString(argName: String, data: Data, isC: Boolean): String = {
+  def getDPICArgString(argName: String, data: Data, isC: Boolean, isFPGA: Boolean = false): String = {
     val typeString = data.getWidth match {
       case 1                                  => if (isC) "uint8_t" else "bit"
       case width if width > 1 && width <= 8   => if (isC) "uint8_t" else "byte"
       case width if width > 8 && width <= 32  => if (isC) "uint32_t" else "int"
       case width if width > 32 && width <= 64 => if (isC) "uint64_t" else "longint"
-      case width if width > 64                => if (isC) "const svBitVecVal" else s"bit[${width - 1}:0]"
+      case width if width > 64 =>
+        if (isFPGA) "uint32_t"
```
klin02, line 47: Change only applys to C side, so maybe change type inside `isC`

xiaokamikami: ok

klin02 (approved):

PR #474, opened 2024-09-24
Batch: pack Batch param to facilitate migration between DPIC/PCIe

klin02:
To facilitate Batch migration between DPIC and PCIe, we pack Batch param to an aligned array, and parse it inside software.

In transmission, we only need to pass single hardware data, as svBitVecVal[] or uint8_t[], and then view it as generated struct in software.

Note we put step inside BatchInfo and call simv_nstep inside, Step will be also exposed to Top all the time for Verilator and Timeout Check.

### DIFF
--- a/src/main/scala/Batch.scala
+++ b/src/main/scala/Batch.scala
@@ -186,7 +186,7 @@ class BatchEndpoint(bundles: Seq[Valid[DifftestBundle]], config: GatewayConfig, 
     }
   }
 
-  val BatchFinish = WireInit(0.U.asTypeOf(new BatchInfo))
+  val BatchFinish = Wire(new BatchInfo)
   BatchFinish.id := (Batch.getTemplate.length + 1).U
   BatchFinish.num := state_step_cnt
   val out = IO(Output(new BatchOutput(chiselTypeOf(state_data), chiselTypeOf(state_info), config)))
--- a/src/main/scala/Batch.scala
+++ b/src/main/scala/Batch.scala
@@ -41,8 +41,8 @@ class BatchIO(dataType: UInt, infoType: UInt) extends Bundle {
   val info = infoType
 }
 
-class BatchOutput(data: UInt, info: UInt, config: GatewayConfig) extends Bundle {
-  val io = new BatchIO(chiselTypeOf(data), chiselTypeOf(info))
+class BatchOutput(dataType: UInt, infoType: UInt, config: GatewayConfig) extends Bundle {
+  val io = new BatchIO(dataType, infoType)
   val enable = Bool()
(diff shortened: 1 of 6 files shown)


src/main/scala/Gateway.scala
```diff
@@ -84,6 +84,7 @@ case class GatewayConfig(
   def check(): Unit = {
     if (hasReplay) require(isSquash)
     if (hasInternalStep) require(isBatch)
+    if (isBatch) require(!hasDutZone)
```
klin02, line 87: In next version of Batch, DiffStates in the same cycle may be splited to different Batch func, so we cannot and no need to pass Batch param to different buffer zones.
To simplify param length division, we disable dutZone with Batch in advance.

src/main/scala/DPIC.scala
```diff
@@ -30,23 +30,28 @@ abstract class DPICBase(config: GatewayConfig) extends ExtModule with HasExtModu
...
   val enable = IO(Input(Bool()))
   val dut_zone = Option.when(config.hasDutZone)(IO(Input(UInt(config.dutZoneWidth.W))))
-  val step = Option.when(config.hasInternalStep)(IO(Input(UInt(config.stepWidth.W))))
 
   def getDirectionString(data: Data): String = {
     if (DataMirror.directionOf(data) == ActualDirection.Input) "input " else "output"
   }
 
-  def getDPICArgString(argName: String, data: Data, isC: Boolean): String = {
+  def getDPICArgString(argName: String, data: Data, isC: Boolean, isDPIC: Boolean = true): String = {
```
klin02, line 38: To support FPGA, which calls v_difftest_Batch inside software and needs `uint8_t io[N]`, we can set isDPIC to false when calling this func to generate dpicFuncProto.

src/main/scala/Batch.scala
```diff
@@ -188,7 +186,10 @@ class BatchEndpoint(bundles: Seq[Valid[DifftestBundle]], config: GatewayConfig,
     }
   }
 
-  val out = IO(Output(new BatchOutput(state_data, state_info, config)))
+  val BatchFinish = WireInit(0.U.asTypeOf(new BatchInfo))
```
poemonsense, line 189: No need to use `WireInit`? `Wire(new BatchInfo)` seems fine

poemonsense (approved):
I don't know all the design details here but the code looks good

src/main/scala/Batch.scala
klin02: I will make a change.

xiaokamikami (commented):
ok

poemonsense (approved):

PR #475, opened 2024-09-25
feat(mem): add optional ref func `ref_memcpy_init`

happy-lx:
When initializing REF's memory, call mem_init function. If REF provides ref_memcpy_init, use it. If not, use ref_memcpy.

### DIFF
--- a/src/test/csrc/difftest/refproxy.h
+++ b/src/test/csrc/difftest/refproxy.h
@@ -135,15 +135,16 @@ public:
   REF_STORE_LOG(f)  \
   REF_DEBUG_MODE(f)
 
-#define REF_OPTIONAL(f)                                                       \
-  f(ref_status, difftest_status, int, )                                       \
-  f(ref_close, difftest_close, void, )                                        \
-  f(ref_set_ramsize, difftest_set_ramsize, void, size_t)                      \
-  f(ref_set_mhartid, difftest_set_mhartid, void, int)                         \
-  f(ref_put_gmaddr, difftest_put_gmaddr, void, void *)                        \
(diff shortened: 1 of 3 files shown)


src/test/csrc/verilator/emu.cpp
```diff
@@ -1064,7 +1064,7 @@ void Emulator::snapshot_save(const char *filename) {
   stream.unbuf_write(&proxy->pc, sizeof(proxy->pc));
 
   char *buf = (char *)mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);
-  proxy->ref_memcpy(PMEM_BASE, buf, size, REF_TO_DUT);
+  proxy->mem_init(PMEM_BASE, buf, size, REF_TO_DUT);
```
poemonsense, line 1067: This line does not need to change because it copies the data from REF to DUT;

src/test/csrc/difftest/refproxy.h
```diff
@@ -268,6 +269,14 @@ class RefProxy : public AbstractRefProxy {
     sync_config();
   }
 
+  inline void mem_init(uint64_t dest, void *src, size_t n, bool direction) {
+    if (ref_memcpy_init) {
+      ref_memcpy_init(dest, src, n, direction);
+    } else {
+      ref_memcpy(dest, src, n, direction);
```
poemonsense, line 276: Should report an error here `#if NUM_CORES > 1`?

After cbo is implemented, ref_memcpy_init must be implemented either in single-core or multi-core.

poemonsense (approved):

poemonsense:
See https://github.com/OpenXiangShan/NEMU/pull/551 for more information

PR #476, opened 2024-09-25
fix(cmo): Support difftest for cbo.inval instruction

happy-lx:
When the DUT executes a cbo.inval, a set is used to record its cacheline address.
Later, if there is a data mismatch between DUT and GoldenMem in the address space operated by the cbo.inval instruction, the Pmem of REF and GoldenMem will be directly updated using the data of DUT.


poemonsense (approved):
The code looks good but I don't know about the `cbo.inval` instruction. Would you mind find another one to review the checker logic?

cailuoshan (approved):

PR #477, opened 2024-09-26
Batch: support data split to reduce transmission bubble

klin02:
Previously, we view data collected from same cycle as a whole, end batch assembling when step data longer than available space. It results in bubble in transmission, and cannot handle situation when step data longer than Max width in a single transmission.

This change support spliting step data according to collector, appending part of data to output and updating remained to state. To shorten logic length, we divide complex logic to three stage.

Note step data may be splited to different batch func, and should be read as a whole, so we avoid buffer-zone switch when batch enabled.

### DIFF
--- a/src/main/scala/Batch.scala
+++ b/src/main/scala/Batch.scala
@@ -24,17 +24,27 @@ import difftest.util.Delayer
 
 import scala.collection.mutable.ListBuffer
 
-case class BatchParam(config: GatewayConfig, dataWidth: Int) {
+case class BatchParam(config: GatewayConfig, collectDataWidth: Int, collectLength: Int) {
   val infoWidth = (new BatchInfo).getWidth
 
+  // Maximum Byte length decided by transmission function
   val MaxDataByteLen = config.batchArgByteLen._1
   val MaxDataBitLen = MaxDataByteLen * 8
-
+  val MaxDataLenWidth = log2Ceil(MaxDataByteLen)
   val MaxInfoByteLen = config.batchArgByteLen._2
   val MaxInfoBitLen = MaxInfoByteLen * 8
+  val MaxInfoLenWidth = log2Ceil(MaxInfoByteLen)
 
-  val BitLenWidth = math.max(log2Ceil(MaxDataBitLen), log2Ceil(dataWidth))
-  val ByteLenWidth = BitLenWidth - 3
+  // Width of statistic for data/info byte length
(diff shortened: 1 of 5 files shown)


klin02:
Tested on Single-core/Dual-core XiangShan. 
Single-core: Calls of Batch reduce by 40%, and simulation speed increases from 450KHz to 490KHz.
Dual-core: Firstly support Batch feature. Simulation speed increases from 140KHz to 200KHz.

After Batch enabled, dual-core Xiangshan can use only single data package(includes DiffState and step trigger) to communicate software, which make it ease to migrate between different platform(such as Palladium, FPGA)

src/main/scala/Batch.scala
```diff
@@ -24,23 +24,39 @@ import difftest.util.Delayer
...
-
+  val MaxDataLenWidth = log2Ceil(MaxDataByteLen)
   val MaxInfoByteLen = config.batchArgByteLen._2
-  val MaxInfoByteWidth = log2Ceil(MaxInfoByteLen)
   val MaxInfoBitLen = MaxInfoByteLen * 8
+  val MaxInfoLenWidth = log2Ceil(MaxInfoByteLen)
+
+  // Width of statistic for data/info byte length
+  val StatsDataLenWidth = log2Ceil((collectDataWidth + 7) / 8)
+  val collectInfoWidth = collectLength * infoWidth
```
poemonsense, line 40: It is better to implement these as `def` instead of `val`. `val` is created every time this case class is intialized, while `def` is only called when necessary. `val` may bring performance issues if this case class is used frequently.

src/main/scala/Batch.scala
```diff
@@ -249,17 +217,197 @@ class BatchCollector(
...
+  val state_info_len = RegInit(0.U(param.MaxInfoLenWidth.W))
+  val state_step_cnt = RegInit(0.U(config.stepWidth.W))
+  val state_trace_size = Option.when(config.hasReplay)(RegInit(0.U(16.W)))
+
+  // Interval update index of buffer, Finish end data parse and call difftest comparision if enabled
+  val BatchInterval = Wire(new BatchInfo)
+  BatchInterval.id := Batch.getTemplate.length.U
+  BatchInterval.num := 0.U // unused
+
+  // Assemble step data/info into state in 3 stage
```
poemonsense, line 258: Comments are good.

poemonsense (dismissed):
I don't know all the design details here. But the code looks good to me.

poemonsense (approved):

PR #482, opened 2024-10-21
difftest: submit buffered data in Squash/Batch when timeout

xiaokamikami:
Commit the remaining cache for squash and batch when the cpu has not committed an instruction for a long time

### DIFF
--- a/src/main/scala/Squash.scala
+++ b/src/main/scala/Squash.scala
@@ -125,6 +125,14 @@ class SquashEndpoint(bundles: Seq[Valid[DifftestBundle]], config: GatewayConfig)
       .reduce(_ || _)
   })
 
+  val timeout_count = RegInit(0.U(32.W))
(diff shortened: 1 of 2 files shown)


src/main/scala/Squash.scala
```diff
@@ -125,6 +125,14 @@ class SquashEndpoint(bundles: Seq[Valid[DifftestBundle]], config: GatewayConfig)
       .reduce(_ || _)
   })
 
+  val time_exceed_count = RegInit(0.U(32.W))
```
poemonsense, line 128: time exceed --> timeout

xiaokamikami: ok

klin02:
When Batch timeout, we should clear state instead of assign new-coming step_data. Logic may be similar to previous state_flush logic, I will try to fix it in this PR

klin02 (approved):

PR #483, opened 2024-10-24
difftest: endpoint connection optimization

xiaokamikami:
When using json to generate difftest, it is necessary to manually connect difftest in ST environment. Now through the generated svh file, it is possible to automatically connect the difftest interface in ST environment without being affected by interface changes


klin02:
We can call generateSvh(numCores) in DifftestTop, and use interfaces in DifftestModule. Seems no need to count numCores again and set config like needSvh

klin02 (approved):

PR #484, opened 2024-10-24
difftest:support direct reading of *.elf workload

zhaoxuying:
from nh3 branch: commit ID 61fb5dec54162389bd17b608a18872cf2f1ddef1  "add sparse memory (support ELF load)"

### DIFF
--- a/src/test/csrc/common/ram.cpp
+++ b/src/test/csrc/common/ram.cpp
@@ -30,6 +30,12 @@
 CoDRAMsim3 *dram = NULL;
 #endif
 
+#ifdef CONFIG_USE_SPARSEMM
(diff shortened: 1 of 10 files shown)


poemonsense (changes requested):
We will not support ELF by implementing the sparsemem and thus will not accept such patches.

ELF files are supported by adding a separate ELF loader, similar to how `gz` and `zstd` files are handled now. Please refer to the source at https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/common/ram.cpp#L279-L287 and https://github.com/OpenXiangShan/difftest/blob/master/src/test/csrc/common/compress.h for more information. You may need to add `elf.h` and `elf.cpp`.

zhaoxuying:
ok, we can close this PR.

PR #485, opened 2024-10-26
feat(ram): Add flat Elf loading capabilities

zarubaf:
This adds the Elf format as an additional pre-loadable file format. Similar to the compressed ones.

### DIFF
--- a/src/test/csrc/common/elfloader.cpp
+++ b/src/test/csrc/common/elfloader.cpp
@@ -15,6 +15,17 @@
 
 #include "elfloader.h"
 
+void ElfBinary::load() {
+  assert(size >= sizeof(Elf64_Ehdr));
+  eh64 = (const Elf64_Ehdr *)raw;
+  assert(IS_ELF32(*eh64) || IS_ELF64(*eh64));
+
+  if (IS_ELF32(*eh64))
+    parse(data32);
+  else
+    parse(data64);
+}
+
 template <typename ehdr_t, typename phdr_t, typename shdr_t, typename sym_t>
 void ElfBinary::parse(ElfBinaryData<ehdr_t, phdr_t, shdr_t, sym_t> &data) {
   data.eh = (const ehdr_t *)raw;
@@ -35,6 +46,25 @@ void ElfBinary::parse(ElfBinaryData<ehdr_t, phdr_t, shdr_t, sym_t> &data) {
       }
     }
   }
+}
+
+ElfBinaryFile::ElfBinaryFile(const char *filename) : filename(filename) {
+  int fd = open(filename, O_RDONLY);
+  struct stat s;
+  assert(fd != -1);
+  assert(fstat(fd, &s) >= 0);
+  size = s.st_size;
+
+  raw = (uint8_t *)mmap(NULL, size, PROT_READ, MAP_PRIVATE, fd, 0);
+  assert(raw != MAP_FAILED);
+  close(fd);
+
+  load();
+}
+
+ElfBinaryFile::~ElfBinaryFile() {
+  if (raw)
+    munmap((void *)raw, size);
 }
 
 bool isElfFile(const char *filename) {
@@ -69,6 +99,12 @@ long readFromElf(void *ptr, const char *file_name, long buf_size) {
 
   uint64_t len_written = 0;
   auto base_addr = elf_file.sections[0].data_dst;
+
+  if (base_addr != PMEM_BASE) {
+    printf(
+        "The first address in the elf does not match the base of the physical memory. It is likely that execution "
+        "leads to unexpected behaviour.");
+  }
 
   for (auto section: elf_file.sections) {
     auto len = section.data_len + section.zero_len;
--- a/src/test/csrc/common/elfloader.cpp
+++ b/src/test/csrc/common/elfloader.cpp
@@ -1,0 +1,84 @@
+/***************************************************************************************
+* Copyright (c) 2024 Axelera AI
+*
+* DiffTest is licensed under Mulan PSL v2.
+* You can use this software according to the terms and conditions of the Mulan PSL v2.
+* You may obtain a copy of Mulan PSL v2 at:
+*          http://license.coscl.org.cn/MulanPSL2
+*
+* THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND,
+* EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+* MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+*
+* See the Mulan PSL v2 for more details.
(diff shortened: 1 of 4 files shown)


poemonsense (commented):
Thanks very much for the work. I think we can close https://github.com/OpenXiangShan/difftest/pull/484 when this PR is merged

1) Are these files written by you and Axelera AI? If some functions are coming from the Internet, we may need to add license or comments there.

2) In some cases, we may want to disable the loading of ELF files, for example, when the raw RISC-V instruction is the same as the magic number in the ELF header. Would you mind add a macro `IMAGE_ELF` with a default value of `1` to the `Makefile`? When the C macro `NO_IMAGE_ELF` is defined, let the `isElfFile` function return `false`. You may refer to https://github.com/OpenXiangShan/difftest/blob/master/Makefile#L154-L160 for a similar macro. If you have other ways to configure it, feel free to let me know.

src/test/csrc/common/elfloader.h
```diff
@@ -0,0 +1,257 @@
...
+  close(fd);
+  return true;
+}
+
+long readFromElf(void *ptr, const char *file_name, long buf_size) {
+    auto elf_file = ElfBinaryFile(file_name);
+    long total_len = 0;
+    for (auto section : elf_file.sections) {
+        size_t section_len = section.data_len + section.zero_len;
+        std::memset(ptr, 0, section_len);
```
poemonsense, line 250: Should this memset be at `ptr + section.data_len` and the size be `section_len - section.data_len`?

src/test/csrc/common/elfloader.h
```diff
@@ -0,0 +1,257 @@
...
+  if (!IS_ELF(*((const Elf64_Ehdr *) buf))) {
+    close(fd);
+    return false;
+  }
+
+  close(fd);
+  return true;
+}
+
+long readFromElf(void *ptr, const char *file_name, long buf_size) {
```
poemonsense, line 245: Since this function is defined in the C/C++ header, should this and `isElfFile` be declared as `static`?

I would recommend moving them (and `ElfBinary::parse`) into a separate `elfloader.cpp` for better code readability.

poemonsense (commented):
Minor code style recommendation.

zarubaf:
Thanks for taking a look!

> Are these files written by you and Axelera AI? If some functions are coming from the Internet, we may need to add license or comments there.

Yes. They are my IP and therefore need to go under Axelera's name.

> In some cases, we may want to disable the loading of ELF files, for example, when the raw RISC-V instruction is the same as the magic number in the ELF header. Would you mind add a macro IMAGE_ELF with a default value of 1 to the Makefile? When the C macro NO_IMAGE_ELF is defined, let the isElfFile function return false. You may refer to https://github.com/OpenXiangShan/difftest/blob/master/Makefile#L154-L160 for a similar macro. If you have other ways to configure it, feel free to let me know.

Done! I think it is best to stay within the same configuration scheme.

I've also made the loading a bit more robust with a couple of sanity checks.

src/test/csrc/common/elfloader.h
zarubaf: Ah yeah that was off, tested it a bit too hastetly when submitting. I've now tried with more sections and fixed the problem. Plus added some more input checks.

poemonsense (dismissed):
The code itself looks good to me now.

But I'm now thinking of another question: how does the ELF address space align with our memory layout? Our RAM has a base address of 0x8000_0000. If the `base_addr` (at line 101) is the same, everything works fine. Otherwise, if the ELF is position dependent, what should be the expected behavior? If the base is higher than 0x8000_0000, we can move it to the base. But what if the base is lower than 0x8000_0000 (most RISC-V user-level ELF has a base at 0x10000)? Should we report a warning in such cases?

@yaozhicheng @Makiras How do you handle this issue in your environment? The base address or memory layout is fixed in the RTL as well as in the simulation framework now. It seems the sparse memory implementation for ELF does not address this issue as well.

src/test/csrc/common/elfloader.cpp
```diff
@@ -0,0 +1,118 @@
...
+long readFromElf(void *ptr, const char *file_name, long buf_size) {
+  auto elf_file = ElfBinaryFile(file_name);
+
+  if (elf_file.sections.size() < 1) {
+    printf("The requested elf '%s' contains zero sections\n", file_name);
+    return -1;
+  }
+
+  uint64_t len_written = 0;
+  auto base_addr = elf_file.sections[0].data_dst;
```
poemonsense, line 101: Probably we are expecting this `base_addr` and the ELF entry point to be 0x8000_0000 now. Unless, the ELF would not execute normally in our simulation platform.

zarubaf:
> But I'm now thinking of another question: how does the ELF address space align with our memory layout? Our RAM has a base address of 0x8000_0000. If the base_addr (at line 101) is the same, everything works fine. Otherwise, if the ELF is position dependent, what should be the expected behavior? If the base is higher than 0x8000_0000, we can move it to the base. But what if the base is lower than 0x8000_0000 (most RISC-V user-level ELF has a base at 0x10000)? Should we report a warning in such cases?

I was thinking the same. Afaik the best we can do is sanity-check the base address with the define `PMEM_BASE` (and issue a warning if it doesn't match)? 

Basically what is happening now is that this implementation blindly copies the content of the elf to the `mem` base address. If the code is position dependent this will of course fail. But you basically run in the same limitation with `objcopy` which I assume you use to convert the elf's to binaries, right?

poemonsense:
> But you basically run in the same limitation with `objcopy` which I assume you use to convert the elf's to binaries, right?

Yes. This is exactly the core reason we didn't implement ELF loader before. The most significant improvement brought by ELF loader is the bin size, which may vary between different workloads.

I think we can wait for the users' feedback on the ELF feature before merging. Maybe tomorrow. Thanks for your patience.

> I was thinking the same. Afaik the best we can do is sanity-check the base address with the define `PMEM_BASE` (and issue a warning if it doesn't match)?

Great if you can help add the warning.

yaozhicheng:
> @yaozhicheng @Makiras How do you handle this issue in your environment? The base address or memory layout is fixed in the RTL as well as in the simulation framework now. It seems the sparse memory implementation for ELF does not address this issue as well.

We did not handle the issue; we think that if the ELF file designates an address, it should be loaded to that corresponding location without alteration. If the program fails to run, the problem should be addressed at its origin, which is where the ELF file is generated.

poemonsense:
> > @yaozhicheng @Makiras How do you handle this issue in your environment? The base address or memory layout is fixed in the RTL as well as in the simulation framework now. It seems the sparse memory implementation for ELF does not address this issue as well.
> 
> We did not handle the issue; we think that if the ELF file designates an address, it should be loaded to that corresponding location without alteration. If the program fails to run, the problem should be addressed at its origin, which is where the ELF file is generated.

Thanks. I think we can merge this PR now.

poemonsense (approved):

poemonsense:
Thanks everyone for contributing to this PR!

PR #486, opened 2024-10-28
feat(dbltrp): add support for critical-error

lewislzh:
This pr raise critical-error in difftest when XiangShan issues a critical-error signal.
The `critical-error` state is defined in the `smdbltrp` extension. This extension specifies that when the `critical-error` signal is asserted, the CPU should halt execution. This PR implements a method for stopping execution in simulation mode, where difftest halts by triggering an `abort`.
In the future, an option to enter debug mode rather than directly halting the CPU will be introduced for handling `critical-error`. This incoming change won’t require modifications to the difftest framework; Xiangshan will only need to control it through an  `valid` signal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ coherence via RefillTest.
 | `DiffLrScEvent` | Executed LR/SC instructions | No |
 | `DiffNonRegInterruptPengingEvent` | Non-register interrupts pending | No |
 | `DiffMhpmeventOverflowEvent` | Mhpmevent-register overflow | No |
+| `DiffDiffCriticalErrorEvent` | Raise critical-error | No |
 
 The DiffTest framework comes with a simulation framework with some top-level IOs.
 They will be automatically created when calling `DifftestModule.finish(cpu: String)`.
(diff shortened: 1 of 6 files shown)


poemonsense (commented):
what's the difference between this and trap?

lewislzh:
> what's the difference between this and trap?
the critical-error state is defined in smdbltrp extension. [critical-error](https://github.com/riscv/riscv-double-trap/blob/main/src/riscv-double-trap.adoc?plain=1#L216-L220)
Xiangshan will send a critical-error signal to the top level and difftest. At this point, the core should stop running: during simulation comparison, difftest will stop by triggering an abort; in other cases, after Xiangshan generates the critical-error signal, the ROB will stop committing instructions.
I think this represents an interruption more severe than exceptions or interrupts, causing the CPU to halt execution directly.

poemonsense:
> In the future, an option will be introduced to enter debug mode in the event of a `critical-error`, allowing the CPU to handle the `critical-error` in a way that improves error diagnosis and resolution. Currently, the handling method for `critical-error` is to directly halt CPU execution.

Make sense. Look good.

lewislzh:
> > In the future, an option will be introduced to enter debug mode in the event of a `critical-error`, allowing the CPU to handle the `critical-error` in a way that improves error diagnosis and resolution. Currently, the handling method for `critical-error` is to directly halt CPU execution.
> 
> Make sense. Look good. Maybe we should add comments to explain its relationship with the double trap extension.

Sure, I have updated the comment to clarify the relationship.

poemonsense (approved):

poemonsense:
Thanks. Merged.

PR #488, opened 2024-10-28
Validate: update ArchRegState only when commit or event

klin02:
When Squash enabled, previous squashed instrCommit may be submitted with new-coming updated ArchRegState, which results in reg mismatch with REF.

In difftest software side, we will only use ArchRegState when instrCommit or event (progress = true in code), so we add updateDependency for ArchRegState to ensure using corresponding ones with squashed commits, just like we add for CSR before.


klin02:
Tested on XiangShan. It works well and fix previous squash-mismatch in some rvv-tests.

poemonsense (approved):

PR #489, opened 2024-10-31
gsim: generate cpp extmodule functions

klin02:
When using GSIM for simulation, we will use generated cpp function to replace extern verilog modules.

This change support generating cppExtFuncs with dependent header, including dpic wrappers and extern module funcs, such as xxHelper.

Note we rename extern module name from DifftestXX to DiffExtXX to avoid extern func name conflicts with stuct. Also, we add uint16_t and shortint support to DPIC args.

### DIFF
--- a/src/main/scala/DPIC.scala
+++ b/src/main/scala/DPIC.scala
@@ -19,7 +19,7 @@ import chisel3._
 import chisel3.experimental.ExtModule
 import chisel3.reflect.DataMirror
 import chisel3.util._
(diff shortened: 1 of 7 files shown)


klin02:
Logic in LogPerfHelper wait for gsim fix chisel annotation and save donTouch top signals.

klin02:
Sequential Logic seems still not supported in externModule cpp-funcs of GSIM. Need further effort.

poemonsense:
Merged in #600

PR #491, opened 2024-10-31
gsim: support gsim based on verilator framework

klin02:


klin02:
Wait PR #489 #490 merged, and need CI with gsim.

src/test/csrc/verilator/emu.cpp
```diff
@@ -663,6 +687,20 @@ inline void Emulator::single_cycle() {
   dut_ptr->coverage_clock = dut_ptr->clock;
 #endif // COVERAGE_PORT_CLOCK
   dut_ptr->eval();
+#endif // VERILATOR
+
+#ifdef GSIM
+  dut_ptr->step();
+  if (dut_ptr->get_difftest$$uart$$out$$valid()) {
```
poemonsense, line 694: Maybe we should improve gsim with better namings? $$ seems strange in the source code. How about proposing changing gsim to be the same as verilator?

klin02: Sure, I will try to propose changing it.

klin02:
fdec03ed656c830c5447c3b6364c8a5
Demo passed with fixed ram(Load bin file to DUT-related variable). Need further support from GSIM.

poemonsense:
Merged by #600

PR #492, opened 2024-11-04
fix(dbltrp): critical-error is not treated as diff error

lewislzh:
* When both RTL and NEMU report a critical-error, the hardware behavior is considered correct. Upon detecting a critical-error, it should indicate a "good trap."
* If they do not report simultaneously, it results in a diff error and triggers an abort.

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -1302,9 +1302,15 @@ void Difftest::do_mhpmevent_overflow() {
 #ifdef CONFIG_DIFFTEST_CRITICALERROREVENT
 void Difftest::do_raise_critical_error() {
   if (dut->critical_error.valid) {
-    display();
-    Info("Core %d dump: critical_error raise \n", this->id);
-    raise_trap(STATE_ABORT);
+    bool ref_critical_error = proxy->raise_critical_error();
+    if (ref_critical_error == dut->critical_error.criticalError) {
+      Info("Core %d dump: critical_error raise \n", this->id);
+      raise_trap(STATE_GOODTRAP);
+    } else {
+      display();
+      Info("Core %d dump: DUT critical_error diff REF \n", this->id);
+      raise_trap(STATE_ABORT);
+    }
   }
 }
 #endif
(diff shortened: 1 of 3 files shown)


src/test/csrc/difftest/refproxy.h
```diff
@@ -101,6 +101,7 @@ class RefProxyConfig {
   f(ref_memcpy, difftest_memcpy, void, uint64_t, void*, size_t, bool)         \
   f(ref_exec, difftest_exec, void, uint64_t)                                  \
   f(ref_reg_display, difftest_display, void, )                                \
+  f(ref_raise_critical_error, difftest_raise_critical_error, void, void*)     \
```
poemonsense, line 104: should be an optional API defined in `REF_OPTIONAL`

src/test/csrc/difftest/difftest.h
```diff
@@ -424,6 +424,7 @@ class Difftest {
 #endif
 #ifdef CONFIG_DIFFTEST_CRITICALERROREVENT
   void do_raise_critical_error();
+  bool ref_critical_error;
```
poemonsense, line 427: No need for this global member.

poemonsense (commented):
Maybe you should declare it as `int ref_raise_critical_error()` and use the return value to check whether it is a legal one.

src/test/csrc/difftest/refproxy.h
lewislzh: Got it, I have fixed it

src/test/csrc/difftest/difftest.h
lewislzh: I have removed this global variable.

src/test/csrc/difftest/difftest.cpp
```diff
@@ -1301,10 +1301,16 @@ void Difftest::do_mhpmevent_overflow() {
 
 #ifdef CONFIG_DIFFTEST_CRITICALERROREVENT
 void Difftest::do_raise_critical_error() {
-  if (dut->critical_error.valid) {
-    display();
-    Info("Core %d dump: critical_error raise \n", this->id);
-    raise_trap(STATE_ABORT);
+  bool ref_critical_error = proxy->raise_critical_error();
+  if (dut->critical_error.valid && dut->critical_error.criticalError) {
```
poemonsense, line 1305: Should it be like:

```c++
if (dut->critical_error.valid && dut->critical_error.criticalError) {
  bool ref_critical_error = proxy->raise_critical_error();
  if (ref_critical_error) ...
  ...
}
```

Only when DUT raises the critical error, we check whether REF raises it?

poemonsense (approved):
Look good to me.

PR #496, opened 2024-11-08
Profile: add public APIs for creating and dumping

poemonsense:
These functions have been implemented before. This commit organizes them together to a separate Scala file.

### DIFF
--- a/src/main/scala/Difftest.scala
+++ b/src/main/scala/Difftest.scala
@@ -21,8 +21,7 @@ import chisel3.reflect.DataMirror
 import chisel3.util._
 import difftest.common.{DifftestWiring, FileControl}
 import difftest.gateway.{Gateway, GatewayConfig}
-import org.json4s.DefaultFormats
-import org.json4s.native.Serialization.writePretty
+import difftest.util.Profile
 
 import scala.annotation.tailrec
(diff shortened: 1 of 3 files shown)


poemonsense:
Currently we simply divide the instances into `{core_0, core_1, ..., core_N}` to get the single-core profile. I'm not sure whether this always works, especially when there is the gatewayendpoint.

src/main/scala/Difftest.scala
```diff
@@ -456,7 +456,7 @@ object DifftestModule {
   private val instances = ListBuffer.empty[DifftestBundle]
   private val cppMacros = ListBuffer.empty[String]
   private val vMacros = ListBuffer.empty[String]
-  private val jsonProfiles = ListBuffer.empty[Map[String, Any]]
+  private val iotrace = ListBuffer.empty[(DifftestBundle, Int)]
```
klin02, line 459: Looks Good to Me. Only some slight question, profile seems no related with trace, maybe ioInterface or some other name will be better?

klin02 (dismissed):

src/main/scala/Difftest.scala
poemonsense: Fixed.

klin02 (approved):

PR #499, opened 2024-11-11
feat(store_event): add pc and robidx debugging information for store event

Anzooooo:


poemonsense:
Should we merge this now?

Anzooooo:
> Should we merge this now?

Yes, thanks.

PR #505, opened 2024-11-13
fix(mem): Fix elf loading byte counting

zarubaf:
I've added a small bug when counting the total bytes written during elf loading. The problem is that we are unpacking the elf and we need to calculate the total size of the resulting memory image, not the "compressed" bytes written.

### DIFF
--- a/src/test/csrc/common/elfloader.cpp
+++ b/src/test/csrc/common/elfloader.cpp
@@ -120,5 +120,8 @@ long readFromElf(void *ptr, const char *file_name, long buf_size) {
     std::memcpy((uint8_t *)ptr + offset, section.data_src, section.data_len);
     len_written += len;
   }
-  return len_written;
+  // Since we are unpacking the sections, the total amount of bytes is the last
+  // section offset plus its size.
+  auto last_section = elf_file.sections.back();
+  return last_section.data_dst - base_addr + last_section.data_len + last_section.zero_len;
 }


src/test/csrc/common/elfloader.cpp
```diff
@@ -120,5 +120,8 @@ long readFromElf(void *ptr, const char *file_name, long buf_size) {
     std::memcpy((uint8_t *)ptr + offset, section.data_src, section.data_len);
     len_written += len;
   }
-  return len_written;
```
poemonsense, line 123: Is this `len_written` still used?

poemonsense (commented):
Look good to me except for the unused `len_written` variable.

After reviewing the code, I noticed there do exist ambiguous meanings for the return value. For the usage in `clone` (ram.h), the return value (as `img_size`) should be the size of max non-zero bytes in the linear address space. I assume the version before this PR is incorrect because the REF will be synced with some data lost.

However, for the usage in `display_stats`, it seems this value is used as the size of meaningful (non-zero) inputs. However, since the stats does not really bring functional issues, I'll fix it in the future and not in this PR.

poemonsense:
Do you have any idea on adding some ELF loading tests?

Much thanks again for your contribution.

poemonsense (approved):
For quick bug fixes, I'm merging this now. We will see how to add elf tests in the next steps.

poemonsense:
Much thanks for your contribution

PR #506, opened 2024-11-15
vcs, palldium: support warmup, set difftest_perfCtrl_clean by DPI-C

xiaokamikami:


klin02:
Previous we use tapAndRead to hidden IO ports for read-only probe like some Difftest Bundle.

However, chisel does not support hierarchical wiring for writable probe, which means even we use tapAndRead, explicit IO port will be generated to pass timer/... By the way, addSink/addSource will also be deprecated later, and latest `bore` API require us to addSource before addSink.

In next step, seems we should still try isVisible to reduce number of LogPerfHelper

klin02:
Seems we can use Builder.currentModule and tap to get LogPerfelper instantiate in same parent Module. I'm trying but some helper instantiated in `when` scope can not accessed from even same module.

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -44,6 +42,8 @@ static bool enable_difftest = true;
 static uint64_t max_instrs = 0;
 static char *workload_list = NULL;
 static uint32_t overwrite_nbytes = 0xe00;
+static uint64_t warmup_instrs = 0;
+static svScope wamupScope;
```
klin02, line 46: `warmup` instead of `wamup`

klin02, line 46: `warmup` instead of `wamup`

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -57,6 +57,8 @@ enum {
   SIMV_FAIL,
 } simv_state;
 
+extern "C" void claer_perfcnt();
```
klin02, line 60: `clear` instead of `claer`

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -84,6 +86,17 @@ extern "C" void set_overwrite_autoset() {
   fclose(fp);
 }
 
+// The workload warms up and clears the instruction counter
+extern "C" void set_warmup_insts(uint64_t warmup_inst, uint64_t cmn_inst) {
```
klin02, line 90: Why cmn_inst here? We only need to pass warmup from vcs_top, and set difftest_clean/dump by dpic

src/test/vsrc/vcs/DifftestEndpoint.v
```diff
@@ -170,12 +185,33 @@ initial begin
...
 
+reg dpi_perfCtrl_clean;
 assign difftest_logCtrl_begin = difftest_logCtrl_begin_r;
 assign difftest_logCtrl_end = difftest_logCtrl_end_r;
 assign difftest_logCtrl_level = 0;
-assign difftest_perfCtrl_clean = 0;
+assign difftest_perfCtrl_clean = dpi_perfCtrl_clean;
 assign difftest_uart_in_ch = 8'hff;
 
+export "DPI-C" task claer_perfcnt;
```
klin02, line 195: MissSpelling here too.

src/test/vsrc/vcs/DifftestEndpoint.v
```diff
@@ -170,12 +185,33 @@ initial begin
...
+task claer_perfcnt();
+  dpi_perfCtrl_clean <= 1'b1;
+endtask
+
+reg clear_perf_cnt;
```
klin02, line 200: Maybe we can simplify logic here. Use RegNext(just example) as mask to ensure `clean` raise only 1 cycle.

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -84,6 +86,17 @@ extern "C" void set_overwrite_autoset() {
...
+  warmup_instrs = warmup_inst + cmn_inst;
```
klin02, line 91: Is it necessary to add CMN here? Difftest is a common verification framework cross mutli-platform, no need to add specific param here if we can add in another way, such as platform-specific scripts.

src/test/vsrc/vcs/DifftestEndpoint.v
```diff
@@ -170,12 +185,23 @@ initial begin
...
-assign difftest_perfCtrl_clean = 0;
+assign difftest_perfCtrl_clean = dpic_perfCtrl_clean;
 assign difftest_uart_in_ch = 8'hff;
 
+export "DPI-C" task clear_perfcnt;
+task clear_perfcnt();
+  dpic_perfCtrl_clean = 1'b1;
+endtask
+
+always @(posedge clock) begin
```
klin02, line 200: Use difftest_perfCtrl_clean_r for same nameing style.

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -84,6 +86,17 @@ extern "C" void set_overwrite_autoset() {
   fclose(fp);
 }
 
+// The workload warms up and clears the instruction counter
+extern "C" void set_warmup_insts(uint64_t warmup_inst) {
+  warmup_instrs = warmup_inst;
+  wamupScope = svGetScope();
```
klin02, line 92: MissSpelling here. `warmup` instead of `wamup`

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -84,6 +86,17 @@ extern "C" void set_overwrite_autoset() {
...
+  if (wamupScope == NULL) {
```
klin02, line 93: No need to check here, but need to check svSetSope.

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -239,11 +252,19 @@ extern "C" uint8_t simv_step() {
       return SIMV_FAIL;
   }
 
-  if (max_instrs != 0) { // 0 for no limit
-    for (int i = 0; i < NUM_CORES; i++) {
+  for (int i = 0; i < NUM_CORES; i++) {
+    auto trap = difftest[i]->get_trap_event();
+    if (warmup_instrs != 0 && trap->instrCnt > warmup_instrs) {
+      svSetScope(wamupScope);
```
klin02, line 258: > Wrap setScope and setSignal to same function. Don't expose Scope to avoid abuse. 
Modified as review before. Like below
```
extern "C" void set_squash_enable(int enable);
void difftest_squash_enable(int enable) {
  if (squashScope == NULL) {
    printf("Error: Could not retrieve squash scope, set first\n");
    assert(squashScope);
  }
  svSetScope(squashScope);
  set_squash_enable(enable);
}
```

src/test/csrc/vcs/vcs_main.cpp
klin02, line 46: Maybe we should seperate scope to warmup. We can add  a context function set_difftest_endpoint_scope and rm `context` from passing warmup_instrs. Later we may also use the scope to change top ports of Difftest like `dump` `clean` and so on.

klin02 (approved):

PR #510, opened 2024-11-21
fix(aia): add support for aia xtopei csr

sinceforYy:
In order for NEMU to use the value of the `xtopi` register for interrupt handling, we need to update the `xtopi` register in NEMU. Since NEMU does not support the out-of-core part of AIA, we need to use the value of `xtopei` when updating `xtopi`, so we pass `xtopei` through the difftest framework.


src/test/csrc/difftest/refproxy.h
```diff
@@ -263,6 +264,12 @@ class RefProxy : public AbstractRefProxy {
     return ref_raise_critical_error ? ref_raise_critical_error() : false;
   }
 
+  inline void aia_xtopei(struct AIAXtopei &xtopei) {
+    if (ref_aia_xtopei) {
+      ref_aia_xtopei(&xtopei);
+    }
```
poemonsense, line 270: Should we print something here to tell the user the API has not been implemented in the REF?

poemonsense (commented):
Look good to me except the warning message when aia is not supported in the REF

poemonsense (approved):

PR #511, opened 2024-11-21
Replay: clear buffered queue, ensure least error report

klin02:
When squashQueue enabled, load/store event may be submitted ahead and buffered in queue before squashed instrs commit along with replay info. So when we recover snapshot and start replaying, we need clear queue, submit and process load/store again.

This change also ensure least report of squashed error before replay, and check if DUT run out of replay range. Also, seperate original step() logic to private do_step(), keep same call interface

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -198,13 +198,14 @@ void Difftest::update_nemuproxy(int coreid, size_t ram_size = 0) {
 #ifdef CONFIG_DIFFTEST_REPLAY
 bool Difftest::can_replay() {
   auto info = dut->trace_info;
-  return !info.in_replay && info.trace_size > 1;
+  return info.valid && !info.in_replay && info.trace_size > 1;
 }
 
 bool Difftest::in_replay_range() {
-  if (dut->trace_info.trace_size > 1)
+  auto info = dut->trace_info;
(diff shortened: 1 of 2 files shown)


klin02:
Tested on XiangShan Palladium. It works well without extra performance loss.

src/test/csrc/difftest/difftest.cpp
```diff
@@ -225,33 +226,65 @@ void Difftest::replay_snapshot() {
...
+    return 0;
+  } else {
+    return ret;
+  }
+#else
+  return do_step();
 #endif // CONFIG_DIFFTEST_REPLAY
+}
+
+inline int Difftest::do_step() {
```
poemonsense, line 286: Maybe we should rename this to like `check_all`?

Others look good to me.

klin02: check_all is also aligned with `check_timeout`, and can make some distinction with some partial check like `do_xx`, so I rename to it.

poemonsense: Maybe we should format the code with better namings in the next step. I'm also considering refactoring the code. The current `difftest.cpp` is too long for us. We need a more modular code structure to implement these checkers.

poemonsense (approved):

src/test/csrc/difftest/difftest.cpp
klin02: I’m also willing to work for it :)

PR #512, opened 2024-11-25
diff: add on fpga batch out io

xiaokamikami:


src/main/scala/util/Profile.scala
```diff
@@ -52,10 +52,8 @@ case class DifftestProfile(
 object DifftestProfile {
   def fromBundles(cpu: String, bundles: Seq[(DifftestBundle, Int)]): DifftestProfile = {
     val numCores = bundles.count(_._1.isUniqueIdentifier)
-    // Note: numCores may be 0 when difftest interfaces are not connected
-    require(numCores == 0 || bundles.length % numCores == 0, "cannot create the profile if cores are not symmetric")
-    val nBundles = if (numCores > 0) bundles.length / numCores else 0
-    val bundleProfiles = bundles.take(nBundles).map { case (b, d) => BundleProfile.fromBundle(b, d) }
+    require(bundles.length % numCores == 0, "cannot create the profile if cores are not symmetric")
```
klin02, line 55: Why change here?

src/main/scala/Gateway.scala
```diff
@@ -234,7 +245,16 @@ class GatewayEndpoint(instanceWithDelay: Seq[(DifftestBundle, Int)], config: Gat
     val batch = Batch(squashed, config)
     step := RegNext(batch.step, 0.U) // expose Batch step to check timeout
     control.enable := batch.enable
-    GatewaySink.batch(Batch.getTemplate, control, batch.io, config)
+    // GatewaySink.batch(Batch.getTemplate, control, batch.io, config)
+
+     val out = IO(Output(new GatewayOutput(batch.io.data, batch.io.info, config)))
```
klin02, line 250: It should be an option IO, and It seems mix Batch logic with common interface.

PR #513, opened 2024-11-25
Batch: expose extra GatewayIO for FPGA

xiaokamikami:
When using FPGA diff, the batch data interface is brought to the gateway top-level
Co-authored-by: zhangshoulin [zzsl1117@gs.zzu.edu.cn](mailto:zzsl1117@gs.zzu.edu.cn)

### DIFF
--- a/src/main/scala/Gateway.scala
+++ b/src/main/scala/Gateway.scala
@@ -53,7 +53,7 @@ case class GatewayConfig(
   def dutBufLen: Int = if (isBatch) batchSize else 1
   def maxStep: Int = if (isBatch) batchSize else 1
   def stepWidth: Int = log2Ceil(maxStep + 1)
-  def batchArgByteLen: (Int, Int) = if (isNonBlock) (3600, 400) else (7200, 800)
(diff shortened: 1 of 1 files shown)


src/main/scala/Gateway.scala
```diff
@@ -234,7 +244,11 @@ class GatewayEndpoint(instanceWithDelay: Seq[(DifftestBundle, Int)], config: Gat
     val batch = Batch(squashed, config)
     step := RegNext(batch.step, 0.U) // expose Batch step to check timeout
     control.enable := batch.enable
-    GatewaySink.batch(Batch.getTemplate, control, batch.io, config)
+    if (!config.isFPGA) {
+      GatewaySink.batch(Batch.getTemplate, control, batch.io, config)
+    } else {
+      GatewaySink.fpga_batch(batch.enable, control, batch.io, config)
```
klin02, line 250: If original GatewaySink.batch is not called, how can you get generated Cpp function like difftest_batch.

klin02 (approved):

PR #514, opened 2024-11-27
Difftest: add the get_all_interfaces API

poemonsense:
This allows the user to get all current interfaces.


poemonsense:
We are adding another `generateJson` method in this PR to allow the user to generate the current json profile. Since DiffTest interfaces are not exposed previously, we add the public method to access it as well.

klin02 (approved):

PR #518, opened 2024-11-28
fix(skip_one): fix floating-point `skip_one` logic

Anzooooo:
Currently, `skip_one` only handles scalar cases. When a floating-point skip_one occurs, the value of the floating-point register is incorrectly synchronized to the scalar register. We have fixed this.
After the modification, we can correctly handle scalar and floating-point skips. However, vector skips will not be allowed, so an assert will be raised when a vector triggers a skip.
Currently, our framework cannot correctly synchronise the data in the vector register. After subsequent discussions, we will add support for this, and this commit will first fix the bug where the floating-point register cannot be correctly synchronised.

---

For reference, the XiangShan processor should never under normal circumstances generate a skip vector, so it is acceptable to add an assert in this place, but we can still choose to add support for vector skip in the future.
The main reason for not adding support for vector skip for the time being is that the width of the vector register needs to be greater than 64, and in some implementations, the vector register may be split into multiple register stacks. For example, in the XiangShan processor, the V0 register is split into a separate register stack.
In summary, we may add support for vector skip in the future, but for now, the priority is to fix the floating-point synchronization bug.

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -527,13 +527,12 @@ int Difftest::do_instr_commit(int i) {
   }
 #endif
 
-  bool realWen = (dut->commit[i].rfwen && dut->commit[i].wdest != 0) || (dut->commit[i].fpwen);
-
   // MMIO accessing should not be a branch or jump, just +2/+4 to get the next pc
   // to skip the checking of an instruction, just copy the reg state to reference design
   if (dut->commit[i].skip || (DEBUG_MODE_SKIP(dut->commit[i].valid, dut->commit[i].pc, dut->commit[i].inst))) {
     // We use the physical register file to get wdata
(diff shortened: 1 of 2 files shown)


poemonsense (approved):
The change itself looks good to me.

However, I need to add a note here for the skip functionality.

As you can see from the code, this a `ref_skip_one` API defined for skipping the execution of an instruction, which has already been defined in Spike. The reason I was adding this API is that, updating the architectural regsiters is not enough for skipping, and updating the performance counters such as `minstret` is also required for more accurate co-simulation.

So, to better support skip for floating-point and vector, actually we also need to implement a `ref_skip_one_v2` in the REF. This is to be done separately in another PR. Both Spike and NEMU are expected to implement it if we are maintaining the correct `minstret` CSR in the REFs (this is important if we are supporting comparing the results of CSRR to `minstret`).

Please let me know if we can merge this now.

Anzooooo:
> minstret

Thanks, this path can be merged.

PR #520, opened 2024-12-03
fix(libso): include headers from RTL simulators

poemonsense:


poemonsense:
To avoid circular dependencies between XiangShan and DiffTest, I'm going to change it to a night test after #529

PR #521, opened 2024-12-09
support perfcct

tastynoob:


tastynoob:
> 1. Please format the commit title, commit message, and code
> 2. Will ChiselDB always generate `perfCCT.cpp`? If not, please use another macro specifically for `perfCCT.cpp`.
> 3. The `main.cpp` currently only applies to Verilator. Should we support `perfCCT.cpp` in VCS?

2:  yes, perfCCT reuse chiselDB init function
3: yes, need some time

tastynoob:
remove global_sim_tick from main.cpp, use verilog dpic function update tick per cycle

poemonsense (commented):
As I still doubt the necessity of introducing `perfCCT.cpp` in ChiselDB, which is expected to create only `chisel_db.cpp`,  the approval of this PR will depend on the approval and merging of https://github.com/OpenXiangShan/Utility/pull/88. Please first consult the maintainers of ChiselDB before changing the ChiselDB C++ APIs in difftest. Thank you

poemonsense (approved):

PR #524, opened 2024-12-25
difftest: Provides a script to change the gatewayendpoint to interface

xiaokamikami:
Provides a script to modify the gatewayendpoint to the interface connection way, for the difftest connection process of the ST environment, and supports the removal of simv top and the removal of simv top from filelist

### DIFF
--- a/scripts/st_tools/interface.py
+++ b/scripts/st_tools/interface.py
@@ -1,0 +1,147 @@
+#/usr/bin/python3
+# -*- coding: UTF-8 -*-
+
+#***************************************************************************************
+# Copyright (c) 2024 Beijing Institute of Open Source Chip (BOSC)
+#
+# DiffTest is licensed under Mulan PSL v2.
+# You can use this software according to the terms and conditions of the Mulan PSL v2.
+# You may obtain a copy of Mulan PSL v2 at:
+#          http://license.coscl.org.cn/MulanPSL2
+#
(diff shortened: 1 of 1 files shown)


poemonsense:
CI fails because of Spike changes. I'll see what's wrong with Spike.

poemonsense:
I'll re-run this after merging https://github.com/OpenXiangShan/riscv-isa-sim/pull/76

scripts/st_tools/interface.py
```diff
@@ -0,0 +1,147 @@
+#/usr/bin/python3
+# -*- coding: UTF-8 -*-
+
+#***************************************************************************************
+# Copyright (c) 2024 Beijing Institute of Open Source Chip (BOSC)
+#
+# XiangShan is licensed under Mulan PSL v2.
```
poemonsense, line 7: DiffTest is licensed ...

poemonsense (commented):
Can you describe how we can use this file in the command line and what it does in this PR? Do we need to add a test for it?

scripts/st_tools/interface.py
xiaokamikami: ok

xiaokamikami:
> Can you describe how we can use this file in the command line and what it does in this PR? Do we need to add a test for it?

I will add the corresponding CI together with adding CI to Xiangshan repository

poemonsense (approved):

PR #525, opened 2024-12-25
feat(ref): add the optional load_flash_bin_v2 API

poemonsense:
This new API accepts the first argument as a pointer to the flash image in memory instead of a string pointer to the flash filename.


poemonsense:
This new API is first introduced in https://github.com/OpenXiangShan/riscv-isa-sim/pull/76.

This change is still compatible with prior versions of NEMU and Spike.

klin02 (approved):

PR #529, opened 2024-12-29
feat(ci): run fuzzing tests daily

poemonsense:
This commit changes the frequency of fuzzing tests in CI to daily.

Running tests for every commit may bring circular dependence between DiffTest and test dependencies. We are currently using the dev-difftest branches in the other repos to avoid this. However, in the future, we may track master branches of the repos.

As fuzzing APIs seldom changes, this functionality does not require tested on each commit. Still, being a nightly test helps us track the changes in xfuzz, rocket, and Spike.


poemonsense:
Merge it now. Let me know if there're any issues

klin02 (approved):

PR #530, opened 2024-12-29
feat(ci): use simpler build commands

poemonsense:


klin02 (approved):

poemonsense:
CI runtime: 
- test-difftest-nutshell: from 18m36s to 16m40s
- test-difftest-vcs: from 14m48s to 13m59s

PR #535, opened 2024-12-30
fix(difftest): wrap cmo.inval code with macros

poemonsense:
This commit fixes the wrong macro that wraps the cmo.inval checkers.


poemonsense:
https://github.com/OpenXiangShan/difftest/pull/476 incorrectly wraps the code in REFILL_EVENT, which may fail if CMO is defined but REFILL is not.

klin02 (approved):

happy-lx (approved):

PR #541, opened 2025-01-03
fix(ci): fix the path for libdifftest.so

poemonsense:


poemonsense:
Tested in https://github.com/OpenXiangShan/difftest/actions/runs/12591238593. All files are correctly uploaded now

klin02 (approved):

PR #552, opened 2025-01-16
fix(DifftestTop): dontTouch DifftestIOs when generated by Profile

klin02:


poemonsense (approved):

klin02:
Previous Difftest works without config H(hierarchical wiring), and miss ports with config H. ( Still has ports in fir, maybe optimized by firtool).
After this change(add dontTouch), Difftest works with config H, but firtool will failed without H. ( We directly read source if visible, left Source Annotation along)

PR #553, opened 2025-01-16
fix(scripts): replace matched interface in original place

klin02:
Previosu this scripts cannot handle port of GatewayEndpoint correctly.

This change restrict replace range inside module ports. And replace target in its original place.


klin02:
Previous we miss some ports in GatewayEndpoint. And we fix that in #552 . However, previous scripts failed to handle these ports.
After this change, scripts act as follows:
image

xiaokamikami (approved):

xiaokamikami (dismissed):

Tang-Haojin (approved):

PR #555, opened 2025-01-20
fix(xdma): remove replicate data parsing when USE_THREAD_MEMPOOL

xiaokamikami:
Remove a temporary fix that was done on the FPGA to fix this issue, and fix a bug where the producer incorrectly parses data in multithreading mode
删除一个临时修复（此问题在FPGA上已修复），修复使用USE_THREAD配置运行时，生产者端进行了重复数据解析的问题

### DIFF
--- a/src/test/csrc/fpga/xdma.cpp
+++ b/src/test/csrc/fpga/xdma.cpp
@@ -138,8 +138,6 @@ void FpgaXdma::stop_thansmit_thread() {
 
 void FpgaXdma::read_xdma_thread(int channel) {
   FpgaPackgeHead *packge = (FpgaPackgeHead *)malloc(sizeof(FpgaPackgeHead));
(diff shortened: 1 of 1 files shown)


src/test/csrc/fpga/xdma.cpp
```diff
@@ -148,13 +146,13 @@ void FpgaXdma::read_xdma_thread(int channel) {
       printf("It should not be the case that no available block can be found\n");
       assert(0);
     }
-#endif // USE_THREAD_MEMPOOL
```
klin02, line 151: if USE_THREAD_MEMPOOL is defined, seems the code will not trigger v_difftest_batch or other func to parse data?

xiaokamikami: The write_difftest_thread function handles parsing when USE_THREAD_MEMPOOL is opened

PR #557, opened 2025-01-31
feat(query): support query for difftest-dpic data

klin02:


klin02:
It also works in Palladium with some viarable set as follows:
export C_INCLUDE_PATH=/path/to/sqlite3/include
export CPLUS_INCLUDE_PATH=/path/to/sqlite3/include
User need to install sqlite3 manually to /path/to/sqlite3.

And in Palladium, after we enable query feature, simulation speed will decrease from 450KHz to 280 KHz (20s -> 30s).

poemonsense (commented):
1) what is this query for? what's the relationship between DPIC and query?

2) should we always generate the query headers?

klin02:
> 1. what is this query for? what's the relationship between DPIC and query?
> 2. should we always generate the query headers?

1. The query function is used to record the data of the DPIC transmission, thus helping to find out optimization strategy of transmission. For example, we can calculate compress ratio of each buffer when incremental transmitting. Also, record of DPIC data helps locating bug in some debugging case.
2. We always generate query headers to avoid generating hardware repeatly for some debugging case, or in some case we will firstly compare performance without Query, then exploring possible reason through Query. This feature will only be enabled when defined CONFIG_DIFFTEST_QUERY.

poemonsense (dismissed):
I don't know much on the functionality of the query but the implementation can be improved.

Functions should be defined in `src` and they will read the predefiend data structures generated by Scala. This is a better way to coordinate between different tools -- they follow some predefined interfaces, especially the C++ structs/classes.

Also, the query C++ functions look similar to ChiselDB. Should we provide some standard DB APIs in DiffTest?

These suggestions are only for future improvements. I don't mind this PR being merged first to help XS development.

klin02:
New Query use BaseClass to wrapper common sqlite behavior, and Scala-generated subClass will only implement some unique behaviour, like get sql Expression and convert structure to sql arg for each Diffstate bundle.

klin02:
Maybe no need to implement subClass, we can pass some sql expression to Class Constructor. I will make a change for that.

PR #558, opened 2025-01-31
Batch: collect from different groups in one cycle to reduce gates

klin02:
Previous we collect data from different groups(a group means Diffstate with same desireCppName) in pipeline (delay different groups to different stage by RegNext). And at each stage Batch will concat valid data from one group to step_data.

This change use two-stage collector to reduce gates: 
Stage1: Different groups will collect data in parallel to get tight-arrangement group_data.
Stage2: Batch will collect data from different groups in one cycle, group_vsize will be used to cut valid data from tight group_data by LookupTree, and valid data from different groups will be concated to step_data.

In pallaium, this change will reduce Batch gates from 21M to 12M. And in FPGA, XS(Basic-Diff) with ESBI config will use 80% LUT (XS needs 70% LUT itself).

### DIFF
--- a/src/main/scala/Batch.scala
+++ b/src/main/scala/Batch.scala
@@ -20,45 +20,49 @@ import chisel3.util._
 import difftest._
 import difftest.gateway.GatewayConfig
 import difftest.common.DifftestPerf
-import difftest.util.Delayer
+import difftest.util.LookupTree
 
 import scala.collection.mutable.ListBuffer
 
-case class BatchParam(config: GatewayConfig, collectDataWidth: Int, collectLength: Int) {
-  def infoWidth = (new BatchInfo).getWidth
+// Only instantiated once, use val instead of def
+case class BatchParam(config: GatewayConfig, bundles: Seq[DifftestBundle]) {
+  val infoWidth = (new BatchInfo).getWidth
+  val infoByte = infoWidth / 8 // byteAligned by default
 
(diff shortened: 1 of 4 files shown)


src/main/scala/Batch.scala
```diff
@@ -20,45 +20,49 @@ import chisel3.util._
...
-import difftest.util.Delayer
+import difftest.util.LookupTree
 
 import scala.collection.mutable.ListBuffer
 
-case class BatchParam(config: GatewayConfig, collectDataWidth: Int, collectLength: Int) {
-  def infoWidth = (new BatchInfo).getWidth
+// Only instantiated once, use val instead of def
+case class BatchParam(config: GatewayConfig, bundles: Seq[DifftestBundle]) {
+  val infoWidth = (new BatchInfo).getWidth
```
poemonsense, line 29: better to use def instead of val

poemonsense (approved):
Don't know the details but assume functionalities are correct

src/main/scala/Batch.scala
klin02: Some params (such as ByteAlignedWidth) are calculated from hardware signals (bundles), use def will make BatchParam illegal to use in other modules. And we will other instantiated BatchParam once to avoid reduplicate calculation.

poemonsense (approved):
I didn't notice the comment in the code. LGTM now.

PR #560, opened 2025-02-01
PerfCnt: count and sum for each DiffState when Batch

klin02:


klin02:
Previous we will only show v_difftest_Batch when enable Batch. This change will also show nums and bytes of every DifftestBundle, which will help find out key DiffState affecting DPIC performance

poemonsense (approved):
We need better ways to define performance counters in the future.

PR #561, opened 2025-02-02
Preprocess: skip loadEvent for single-core

klin02:
After optimized by Squash, DPIC bytes of LoadEvent accounts for nearly 30% of total. However in do_load_check(difftest.cpp), LoadEvent will only be used when numCores greater than 1, so we skip LoadEvent for singleCore.

### DIFF
--- a/src/main/scala/Preprocess.scala
+++ b/src/main/scala/Preprocess.scala
@@ -81,7 +81,16 @@ class PreprocessEndpoint(bundles: Seq[DifftestBundle]) extends Module {
     cd
   }
 
-  val withCommitData = MixedVecInit((in.filterNot(_.desiredCppName.contains("wb")) ++ commitData).toSeq)
-  val out = IO(Output(chiselTypeOf(withCommitData)))
-  out := withCommitData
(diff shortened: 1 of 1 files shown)


klin02:
BTW, according to DPIC data recorded by Query, Increment transmitting works for other bundle taking up major part of DPIC, including RegState/CSR/RefillEvent, but not LoadEvent. 
LoadEvent accounts for a large part of DPIC, but will not be checked for single-core. We skip LoadEvent to highlight the benefits of Incremental.

poemonsense (approved):
We may also disable DPI-C calls of LoadEvent in single-core configs (possibly by adding ifdef NUM_CORES in DPI-C functions?)

klin02:
> We may also disable DPI-C calls of LoadEvent in single-core configs (possibly by adding ifdef NUM_CORES in DPI-C functions?)

Maybe just leave DiffLoadEvent ungenerated for single-core? 
Now with Preprocess we will skip LoadEvent if we use any difftest-config(then we will use GatewayEndpoint), shall we also skip it by default (now used in common verilator flow)?

PR #562, opened 2025-02-06
Batch: disable split strategy for FPGA to reduce gates

klin02:
Previous we enable splitting Batch data to handle peak data valid in same cycle (we call these data collected in same cycle Stepdata). The step data may be splitting according to remain space of Batch state, and part of them will be appended to Batch output to make full use of Batch DPIC.

However, this split strategy need some gates to handle splitting and appending behaviour. Now for FPGA, we need to reduce gates of Batch, so we disable split strategy.

According to gateCount of Palladium, disable split strategy will reduce gates of Batch from 12M to 8M (in full Difftest). And in FPGA, LUT of XS and Difftest(Basic-diff) will reduce from 81% to 77%, with XS accounts for 70% itself.

### DIFF
--- a/src/main/scala/Batch.scala
+++ b/src/main/scala/Batch.scala
@@ -48,7 +48,6 @@ case class BatchParam(config: GatewayConfig, bundles: Seq[DifftestBundle]) {
 
   // Truncate width when shifting to reduce useless gates
   val TruncDataBitLen = math.min(MaxDataBitLen, StepDataBitLen)
-  val TruncInfoBitLen = math.min(MaxInfoBitLen, StepInfoBitLen)
 }
 
 class BatchIO(param: BatchParam) extends Bundle {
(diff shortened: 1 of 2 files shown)


src/main/scala/Batch.scala
```diff
@@ -230,46 +229,64 @@ class BatchAssembler(
...
-    }).reduceTree(_ | _).asTypeOf(new BatchStats(param)),
-  )
-  val remain_stats = Wire(new BatchStats(param))
-  remain_stats.data_bytes := delay_step_status.last.data_bytes -& concat_stats.data_bytes
-  remain_stats.info_size := delay_step_status.last.info_size -& concat_stats.info_size
-  assert(remain_stats.data_bytes <= param.MaxDataByteLen.U)
-  assert(remain_stats.info_size + 1.U <= param.MaxInfoSize.U)
+  val data_exceed = Wire(Bool())
+  val info_exceed = Wire(Bool())
+  val step_all_exceed = Option.when(config.batchSplit)(Wire(Bool()))
```
poemonsense, line 234: For so many `Option.when`, it is better to use

```
if (config.batchSplit) {
}
else {
...
}
```

src/main/scala/Batch.scala
```diff
@@ -286,83 +303,102 @@ class BatchAssembler(
...
   BatchFinish.id := (Batch.getTemplate.length + 1).U
   BatchFinish.num := finish_step
   // Use BatchInterval to update index of software buffer
   val BatchInterval = Wire(new BatchInfo)
   BatchInterval.id := Batch.getTemplate.length.U
   BatchInterval.num := delay_step_status.last.info_size // unused, only for debugging
-  val append_finish_map = Seq.tabulate(param.StepGroupSize) { g =>
-    (g.U, (BatchFinish.asUInt << (g * param.infoWidth)).asUInt)
+
+  val append_data = if (config.batchSplit) {
```
poemonsense, line 322: this is the same. Readability can be improved

poemonsense (approved):
Some minor style improvements.

Maybe we should let @xiaokamikami known more on the functionalities and help review them in the future. It seems difficult for me to know all details because I don't participate in daily discussions.

xiaokamikami:
> Some minor style improvements.
> 
> Maybe we should let @xiaokamikami known more on the functionalities and help review them in the future. It seems difficult for me to know all details because I don't participate in daily discussions.

I have no objections to this approach. Regarding the FPGA, reducing resource utilization is beneficial, and the data configuration can be fine-tuned based on the available PCIe bandwidth.

src/main/scala/Batch.scala
klin02: As we seperate logic of data split and state update, some signals can only be declared as Option. If we wrap them in if/else, it seems hard to cover some shared logic in same if/else.

klin02:
This change wrap all middle variable inside if/else to avoid too many Options.

poemonsense (approved):

PR #563, opened 2025-02-06
vcs: Ensure correct execution order of difftest DPI calls

forever043:
An additional `simv_step_event` was introduced to control the execution order of the `simv_nstep()` DPI call. The order of DPI calls in different `always` blocks is inherently unpredictable. Since `simv_nstep()` depends on the state updated by other `v_difftest_*` DPI calls, a `#0.1` delay was added to ensure that `simv_nstep()` is executed at the end of each clock posedge. This modification resolves potential timing issues caused by the implicit dependency between difftest DPI calls.

Additionally, the file extension of `DifftestEndpoint.v` was changed to `.sv`, as the `event` mechanism is a feature specific to SystemVerilog.

### DIFF
--- a/src/test/vsrc/vcs/DifftestEndpoint.sv
+++ b/src/test/vsrc/vcs/DifftestEndpoint.sv
@@ -269,6 +269,13 @@ DeferredControl deferred(
 );
 `else
 reg [7:0] simv_result;
+`ifdef PALLADIUM
+/*
+ * In PALLADIUM, we delay the step signal to next cycle to make sure
+ * `simv_step` was triggered after other DPI calls, which needs more
+ * trick to be correct. Such as introducing `ping-pong buffer` to
+ * handle the delay-step, and dpics at the next cycle coming together.
+ */
 always @(posedge clock) begin
   if (reset || simv_result == `SIMV_DONE) begin
     simv_result <= 8'b0;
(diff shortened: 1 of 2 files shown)


forever043:
we found the `simv_nstep()` was called before `v_difftest_FpCSRState`  update `dut->fcsr` in our VCS testbench, which cause difftest check failed if instr commit update `fcsr`.

poemonsense (commented):
I remembered that we did discuss this issue before @klin02 

Updates to DiffState should take effects before `step`. I think this was achieved by updating it at the previous clock cycle. Is it correct?

klin02:
https://github.com/OpenXiangShan/difftest/blob/master/src%2Fmain%2Fscala%2FGateway.scala#L242-L242

We have considered this issue before, see details above. We delay step signal to make sure simv_nstep called after dpic. Note this delay will only be applyed with GatewayEndpoint, which means we should also set difftest-config Z for vcs flow.

 See more details in Difftest CI

src/test/vsrc/vcs/DifftestEndpoint.sv
```diff
@@ -269,6 +269,13 @@ DeferredControl deferred(
 );
 `else
 reg [7:0] simv_result;
+`ifdef PALLADIUM
+/*
+ * CAUTION: The DPI-C execution order was not handled correctly,
+ * for PALLADIUM platform it is possible that the difftest step
+ * was executed **BEFORE** all dffftest state update finished.
+ * (fortunately, it was not happened yet)
+ */
```
klin02, line 278: In palladium, we will delay step signal to next cycle, to make sure simv_step trigger after other dpic. That need more trick to be correct, like a ping-pong buffer to handle delay-step and dpics at the next cycle coming together.

klin02 (approved):

src/test/vsrc/vcs/DifftestEndpoint.sv
forever043: fixed

```
/*
 * In PALLADIUM, we delay the step signal to next cycle to make sure
 * `simv_step` was triggered after other DPI calls, which needs more
 * trick to be correct. Such as introducing `ping-pong buffer` to
 * handle the delay-step, and dpics at the next cycle coming together.
 */
```

klin02 (approved):

PR #564, opened 2025-02-07
Batch: rename BatchInterval to BatchStep, move to tail of StepInfo

klin02:
Previous we append BatchInterval at the head of StepInfo, using it to update index of software buffer.
This change rename BatchInterval to BatchStep, and append it to StepInfo in BatchCollector rather than BatchAssembler to simplify logic. Also, putting BatchStep at tail of StepInfo allow some operation after Batch parsing, which is needed by Incremental
transmit.


klin02:
This PR is based on #562 . We split it to different PR to claify code changes, and will rebase this PR after merging another.

xiaokamikami (approved):

xiaokamikami (approved):

PR #566, opened 2025-02-09
fix(elfloader): Sort phdr entries by paddr before return to `readFromElf()`

forever043:
The `readFromElf()` relies on all phdr entries are sorted by paddr, which is not mandatory for the ELF format. We must sort the parsed `sections` before return back to `readFromElf()`.


poemonsense (approved):
First time I know this elf detail

poemonsense (approved):

poemonsense:
should we merge this now?

forever043:
> First time I know this elf detail

ready-to-merge :)

PR #571, opened 2025-02-14
fix(Makefile): Link libdramsim3.a by `-L/-l` instead of absolute path

forever043:


poemonsense (approved):
what's the difference? I don't remember why we simply add the `.a` instead of using `-L`

forever043:
> what's the difference? I don't remember why we simply add the `.a` instead of using `-L`

maybe you want to link the static lib explictly?

in my case (libso.mk), the `SIM_LDFLAGS` was merged into `LIB_CXXFLAGS`, which was used to compile (w/ gcc -c) all *.cc sources.
if we use the absolate path, the final link cmd will try to link with the absolate path even it was not placed in the link objects（<--but why?)

unfortunately, my libdramsim3.a was not compiled into the `$(DRAMSIM3_HOME)/build` directory, the `difftest-so` target report link error of not find the `$(DRAMSIM3_HOME)/build/libdramsim3.a`.

because `gcc -c` will ignore all `link-related options(a.k.a. -L/-l)`, this fix may save my environment a little bit :)

poemonsense:
> > what's the difference? I don't remember why we simply add the `.a` instead of using `-L`
> 
> maybe you want to link the static lib explictly?
> 
> in my case (libso.mk), the `SIM_LDFLAGS` was merged into `LIB_CXXFLAGS`, which was used to compile (w/ gcc -c) all *.cc sources. if we use the absolate path, the final link cmd will try to link with the absolate path even it was not placed in the link objects（<--but why?)
> 
> unfortunately, my libdramsim3.a was not compiled into the `$(DRAMSIM3_HOME)/build` directory, the `difftest-so` target report link error of not find the `$(DRAMSIM3_HOME)/build/libdramsim3.a`.
> 
> because `gcc -c` will ignore all `link-related options(a.k.a. -L/-l)`, this fix may save my environment a little bit :)

Did you test whether XiangShan with DRAMsim3 compiles correctly? If so, I think we are good to merge this

forever043:
> Did you test whether XiangShan with DRAMsim3 compiles correctly? If so, I think we are good to merge this

test ok with:
```
make emu WITH_DRAMSIM3=1
```

image

poemonsense:
Thanks for the work

PR #578, opened 2025-02-21
vcs: soft warmup support and more ipc info dump

xiaokamikami:
To retain the warmup function even when the performance counter is not enabled, a software-based counting method has been added. Meanwhile, when printing IPC, the data without warmup (original IPC) is still retained for printing.


Makefile
```diff
@@ -85,6 +85,9 @@ WITH_CHISELDB = 0
 WITH_CONSTANTIN = 0
 endif
 
+ifeq ($(ENABLE_CORE_PERF_COUNT),1)
+SIM_CXXFLAGS += -DENABLE_CORE_PERF_COUNT
```
poemonsense, line 89: Do we need this macro?

1) if +warmup_instr is given, we show warmup_ipc and sampling_ipc
2) otherwise, we should IPC (as the previous one)

xiaokamikami: If we don't add this macro, which uses +warmup_instr by default, when warmup is triggered it will try to empty the DUT's hardware performance counter, if we don't turn on the performance counter, I'm not sure if this will cause some problems

poemonsense: It does not matter. Calling `difftest_perfCtrl_clean` only set some top-level signals. Even if the DUT does not have any performance counters, the simulation still works. For example, the CI is using NutShell, but NutShell does not have any performance counters.

poemonsense (approved):

PR #586, opened 2025-02-26
fix: Save the clock and instruction count for the warmup

xiaokamikami:
The clock and instruction counter data should be saved at the time of the warmup to calculate the IPC after the warmup，It also supports recording the warmup data of each core when multiple cores are used，and change the command print prompt when the warmup trigger is triggered to base 10


poemonsense (dismissed):

src/test/csrc/vcs/vcs_main.cpp
```diff
@@ -283,6 +285,11 @@ extern "C" uint8_t simv_step() {
       }
     }
   }
+
+  if (need_warmup) {
+    need_warmup = false;
```
poemonsense, line 290: This line can be removed. Others looks good to me

poemonsense (approved):

PR #588, opened 2025-02-26
Batch: adopt cluster and shifter for better area and timing

klin02:
Previous we collect data from different group by MuxTree and Cat, however it leads to bad timing for multi-level mux, and consuming more gates if we insert some register to optimize timing.

This change use three stage batch for better area and timing: 

Stage 1: we use BatchCluster to cluster data with same class in same cycle, which means make valid data close to each other. As each bundle has same width, we can simplify use mux to choose location. Note we also insert a RegNext before cluster to cut off combinatorial logic circuit.

Stage 2: we use BatchCollector to collect data with different class, but also in same cycle. As each group_data has different width and different valid bytes, we use shift to collect group_data instead of original muxtree for better area and timing.

Stage 3: we use BatchAssembler to assemble data from different cycles. Assembler will split step_data according to available remaining space, to make full use Batch DPIC.

### DIFF
--- a/src/main/scala/Batch.scala
+++ b/src/main/scala/Batch.scala
@@ -94,7 +94,7 @@ class BatchEndpoint(bundles: Seq[Valid[DifftestBundle]], config: GatewayConfig) 
 
   // Collect valid bundles of same cycle
   val collector = Module(new BatchCollector(bundles, param))
-  collector.in := in
+  collector.in := RegNext(in)
   val step_data = collector.step_data
   val step_info = collector.step_info
   val step_enable = collector.step_enable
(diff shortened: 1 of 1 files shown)


klin02:
Tested in Palladium, Gates of Batch will reduce from 18 M to 10 M with new feature Delta (not merged yet). 
And steps (means logic circuit length) reduce from 500+ to 330.

klin02:
Wait until tested in FPGA

klin02:
> Wait until tested in FPGA

poemonsense (approved):

PR #590, opened 2025-02-27
fix: fix missing header file and mpool copy width

xiaokamikami:
Given that the data width in difftest varies across different configurations, it is necessary to request mempool to generate memory blocks according to the specific data width and regulate the width of copied data accordingly. 
Additionally, DIFFTEST-DPIC.h must be included in xdma.cpp to enable the invocation of dpic processing functions.


klin02 (dismissed):

xiaokamikami:
I encapsulate the basic block of the memory pool as a class so that it can be instantiated to different sizes in different places

xiaokamikami:
I encapsulate the basic block of the memory pool as a class so that it can be instantiated to different sizes in different places

poemonsense (approved):
Some init values look unnecessary but overall looks good to me

PR #592, opened 2025-02-28
FPGA: exclude unused svdpi.h

xiaokamikami:


klin02 (dismissed):

src/main/scala/DPIC.scala
```diff
@@ -357,9 +357,11 @@ object DPIC {
     interfaceCpp += ""
     interfaceCpp += "#include <cstdint>"
     interfaceCpp += "#include \"diffstate.h\""
+    interfaceCpp += "#ifndef CONFIG_PLATFORM_FPGA"
```
klin02, line 360: Or we can use define like follows:
#if defined(CONFIG_DIFFTEST_BATCH) && !defined(CONFIG_PLATFORM_FPGA)
may be more clear then nested ifdef

PR #593, opened 2025-02-28
Delta: support bundle spliting and filter duplicate elem

klin02:
This change add Delta feature to speed up difftest in Palladium/FPGA. We enable spliting bundle into elems, and only transmit non-duplicate elem through DPIC.

As we need a buffer to record base value for delta-transmission, we generate difftest-delta.h, which provides API to update delta data from elem's DPIC, and sync final data to diffState (just like non-delta transmission). Currectly we rely on Batch to ensure Delta order, which updates delta-buffer during DPIC parsing, and sync at the end of each step.

Based on Config ESBIN (enable Squash and Batch），Delta feature will reduce DPIC bytes by ~40%，which will be better without Squash or under Basic-Diff, where duplicate data from bundle like ArchReg/CSR... acounts for more proportions of DPIC bytes.

### DIFF
--- a/src/main/scala/DPIC.scala
+++ b/src/main/scala/DPIC.scala
@@ -22,6 +22,7 @@ import chisel3.util._
 import difftest._
 import difftest.batch.{BatchInfo, BatchIO}
 import difftest.common.FileControl
+import difftest.delta.Delta
 import difftest.gateway.{GatewayConfig, GatewayResult, GatewaySinkControl}
 import difftest.util.Query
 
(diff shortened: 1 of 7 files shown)


src/main/scala/DPIC.scala
```diff
@@ -207,12 +212,17 @@ class DPICBatch(template: Seq[DifftestBundle], batchIO: BatchIO, config: Gateway
...
+    } else {
+      s"sizeof(${gen.desiredModuleName})"
+    }
+    unpack += s"memcpy(packet, data, $size);"
     unpack += s"data += ${elem_bytes.sum};"
     unpack +=
       s"""
          |#ifdef CONFIG_DIFFTEST_QUERY
-         |  ${Query.writeInvoke(gen)}
+         |        ${Query.writeInvoke(gen)}
```
poemonsense, line 225: fix this

poemonsense (approved):
We need to refactor the software structure in the future for, like, difference optimization passes, backends, ...

src/main/scala/DPIC.scala
klin02: Sure, I will try to refactor in the future.
And this change will modify index like follows:
image

poemonsense: so this does not need to be changed? Then we can merge this PR

klin02: Yes. We can merge this PR

PR #594, opened 2025-03-03
feat(load_check): add vec load check

Anzooooo:
This commit adds vector load check.
This is mostly done in the following way:
1. checking for consistency of vector load results in dut and ref's pmem.
2. if not, then check if the vector load results in dut and golden mem are consistent.
3. if the check in (2) passes, update ref.

---

Consider the complexity and diversity of vector.
While accessing pmem in ref, a golden mem access will be performed at the same time, and the related information (addr, data) will be stored temporarily.
When you need to compare the golden mem, you will get the golden mem access result from ref via API.
When the golden mem needs to be updated, the API will ask ref to get the information (addr and data) from the stored information and write them to the golden mem.

---

**I've tested this in a number of different emulations and have added adaptation code for different acceleration strategies.
After testing, this will bring some extra resource usage on palladium, probably around 800W. I'm not sure if the size of this increase is what is expected, so please confirm this for me.**

### DIFF
--- a/src/main/scala/Bundles.scala
+++ b/src/main/scala/Bundles.scala
@@ -63,8 +63,11 @@ class InstrCommit(val numPhyRegs: Int = 32) extends DifftestBaseBundle with HasV
   val rfwen = Bool()
   val fpwen = Bool()
   val vecwen = Bool()
+  val v0wen = Bool()
   val wpdest = UInt(log2Ceil(numPhyRegs).W)
   val wdest = UInt(8.W)
+  val otherwpdest = Vec(8, UInt(8.W))
+  val otherV0Wen = Vec(8, Bool())
 
   val pc = UInt(64.W)
   val instr = UInt(32.W)
@@ -87,6 +90,7 @@ class InstrCommit(val numPhyRegs: Int = 32) extends DifftestBaseBundle with HasV
 // Instantiate inside DiffTest, work for get_commit_data specially
 class CommitData extends DifftestBaseBundle with HasValid {
   val data = UInt(64.W)
+  val vecData = Vec(16, UInt(64.W))
 }
 
 class TrapEvent extends DifftestBaseBundle {
@@ -166,6 +170,10 @@ class DataWriteback(val numElements: Int) extends DifftestBaseBundle with HasVal
   val data = UInt(64.W)
 }
 
+class VecDataWriteback(val numElements: Int) extends DifftestBaseBundle with HasValid with HasAddress {
+  val data = Vec(2, UInt(64.W))
+}
+
 class ArchIntRegState extends DifftestBaseBundle {
   val value = Vec(32, UInt(64.W))
 
@@ -234,7 +242,8 @@ class LoadEvent extends DifftestBaseBundle with HasValid {
   val paddr = UInt(64.W)
   val opType = UInt(8.W)
   val isAtomic = Bool()
-  val isLoad = Bool() // Todo: support vector load
+  val isLoad = Bool()
+  val isVLoad = Bool()
 }
 
 class AtomicEvent extends DifftestBaseBundle with HasValid {
--- a/src/main/scala/Difftest.scala
+++ b/src/main/scala/Difftest.scala
@@ -297,8 +297,15 @@ class DiffFpWriteback(numRegs: Int = 32) extends DiffIntWriteback(numRegs) {
   override val desiredCppName: String = "wb_fp"
 }
 
-class DiffVecWriteback(numRegs: Int = 32) extends DiffIntWriteback(numRegs) {
+class DiffVecWriteback(numRegs: Int = 32) extends VecDataWriteback(numRegs) with DifftestBundle {
   override val desiredCppName: String = "wb_vec"
+  override protected val needFlatten: Boolean = true
+  override def supportsSquashBase: Bool = true.B
+  override def classArgs: Map[String, Any] = Map("numRegs" -> numRegs)
+}
+
+class DiffVecV0Writeback(numRegs: Int = 32) extends DiffVecWriteback(numRegs) {
+  override val desiredCppName: String = "wb_v0"
 }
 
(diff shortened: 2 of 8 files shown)


src/main/scala/Bundles.scala
```diff
@@ -235,6 +243,7 @@ class LoadEvent extends DifftestBaseBundle with HasValid {
   val opType = UInt(8.W)
   val isAtomic = Bool()
   val isLoad = Bool() // Todo: support vector load
+  val isVLoad = Bool() // Todo: support vector load
```
poemonsense, line 246: remove todo here

src/test/csrc/difftest/refproxy.h
```diff
@@ -153,7 +153,10 @@ class RefProxyConfig {
   f(ref_raise_critical_error, difftest_raise_critical_error, bool)                                          \
   f(ref_get_store_event_other_info, difftest_get_store_event_other_info, void, void*)                       \
   f(ref_sync_aia, difftest_sync_aia, void, void*)                                                           \
-  f(ref_sync_custom_mflushpwr, difftest_sync_custom_mflushpwr, void, bool)
+  f(ref_sync_custom_mflushpwr, difftest_sync_custom_mflushpwr, void, bool)                                  \
+  f(ref_get_vec_load_vdNum, difftest_get_vec_load_vdNum, int, )                                                 \
+  f(ref_get_vec_load_dual_goldenmem_reg, difftest_get_vec_load_dual_goldenmem_reg, void*, )                                                       \
```
poemonsense, line 158: What are these new functions for? What's the relationship between this and the original goldenmem?

poemonsense (commented):
It's surprising you have done a lot of work.

I don't fully understand the details here. Why do we need these new APIs for vec load? Generally speaking, I'm viewing vectored/scalar loads as N/1 64-bit memory loads, and don't understand the fundamental reason that we need such special handling logic for vector?

src/main/scala/Difftest.scala
```diff
@@ -297,8 +297,17 @@ class DiffFpWriteback(numRegs: Int = 32) extends DiffIntWriteback(numRegs) {
...
   override val desiredCppName: String = "wb_vec"
+  override protected val needFlatten: Boolean = true
+  override def supportsSquashBase: Bool = true.B
+  override def classArgs: Map[String, Any] = Map("numRegs" -> numRegs)
+  override val data: Vec[UInt] = Vec(2, UInt(64.W))
+}
+
+class DiffVecV0Writeback(numRegs: Int = 32) extends DiffVecWriteback(numRegs) {
+  override val desiredCppName: String = "wb_v0"
+  override val data: Vec[UInt] = Vec(2, UInt(64.W))
```
klin02, line 310: Why we need to override `data` in BaseBundle, and I'm not quite sure what's the difference between Vec and VecV0

klin02:
Seems the basic flow/idea of Vec load is same as origin load? Can we give a common function or API to them, just like StoreEvent.( Of course, may be some elem is needed for this common bundle, to distinguish between vec-load or load）

src/main/scala/Difftest.scala
Anzooooo: Currently, the xiangshan implementation has vec and v0 as two separate register stacks.
Either choose to rewrite a new wb_v0 or instantiate two wb_vec.
As previously written, I chose to replicate a new wb_v0.

---

The implementation in XiangShan is as follows:

```Scala
  // difftest
  if (env.EnableDifftest || env.AlwaysBasicDiff) {
    intWbArbiterOut.foreach(out => {
      val difftest = DifftestModule(new DiffIntWriteback(IntPhyRegs))
      difftest.coreid := io.fromTop.hartId
      difftest.valid := out.fire && out.bits.rfWen
      difftest.address := out.bits.pdest
      difftest.data := out.bits.data
    })
  }

  if (env.EnableDifftest || env.AlwaysBasicDiff) {
    fpWbArbiterOut.foreach(out => {
      val difftest = DifftestModule(new DiffFpWriteback(FpPhyRegs))
      difftest.coreid := io.fromTop.hartId
      difftest.valid := out.fire // all fp instr will write fp rf
      difftest.address := out.bits.pdest
      difftest.data := out.bits.data
    })
  }

  if (env.EnableDifftest || env.AlwaysBasicDiff) {
    vfWbArbiterOut.foreach(out => {
      val difftest = DifftestModule(new DiffVecWriteback(VfPhyRegs))
      difftest.coreid := io.fromTop.hartId
      difftest.valid := out.fire
      difftest.address := out.bits.pdest
      difftest.data(0) := out.bits.data(63, 0)
      difftest.data(1) := out.bits.data(127, 64)
    })
  }

  if (env.EnableDifftest || env.AlwaysBasicDiff) {
    v0WbArbiterOut.foreach(out => {
      val difftest = DifftestModule(new DiffVecV0Writeback(V0PhyRegs))
      difftest.coreid := io.fromTop.hartId
      difftest.valid := out.fire
      difftest.address := out.bits.pdest
      difftest.data(0) := out.bits.data(63, 0)
      difftest.data(1) := out.bits.data(127, 64)
    })
  }
```

poemonsense (approved):
ok to merge this? @klin02

klin02:
> ok to merge this? @klin02

I have some minor suggestion to @Anzooooo , and the performance/area in Palladium is eavluating now.

Anzooooo:
> > ok to merge this? @klin02
> 
> I have some minor suggestion to @Anzooooo , and the performance/area in Palladium is eavluating now.

Yes, I'm running some palladium emulation performance tests, which will take a while due to resource constraints, and expect to have detailed results by this afternoon.

klin02 (approved):

poemonsense:
Let me know if this is ready for merging please

Anzooooo:
> Let me know if this is ready for merging please

Yes, it's ready on my end, if there are no other questions, please merge the pr.
I will update XiangShan with NEMU after the merge.

PR #599, opened 2025-03-12
feat(difftest): add output optional parameter for dramsim3

zephyrols:
Add parameter --dramsim3-outdir

### DIFF
--- a/src/test/csrc/verilator/emu.cpp
+++ b/src/test/csrc/verilator/emu.cpp
@@ -109,6 +109,7 @@ static inline void print_help(const char *file) {
   printf("      --remote-jtag-port     specify remote bitbang port\n");
 #ifdef WITH_DRAMSIM3
   printf("      --dramsim3-ini         specify the ini file for DRAMSim3\n");
+  printf("      --dramsim3-outdir      specify the output dir for DRAMSim3\n");
 #endif
 #if VM_COVERAGE == 1
   printf("      --dump-coverage        enable coverage dump\n");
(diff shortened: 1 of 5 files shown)


src/test/csrc/verilator/emu.cpp
```diff
@@ -97,6 +97,7 @@ static inline void print_help(const char *file) {
 #ifdef ENABLE_CHISEL_DB
   printf("      --dump-db              enable database dump\n");
   printf("      --dump-select-db       select database's table to dump\n");
+  printf("      --db-path              the directory used to store the database\n");
```
poemonsense, line 100: Could you please separate this argument from this PR?

It is better to have separate PRs for dramsim3 and ChiselDB

src/test/csrc/verilator/emu.cpp
```diff
@@ -145,6 +147,7 @@ inline EmuArgs parse_args(int argc, const char *argv[]) {
     { "sim-run-ahead",     0, NULL,  0  },
     { "dump-db",           0, NULL,  0  },
     { "dump-select-db",    1, NULL,  0  },
+    { "db-path",           1, NULL,  0  },
```
poemonsense, line 150: Since this struct can differ from the `--help` strings, there's no need to add this argument here. You can simply append this argument at the end of the argument list. Then there won't be such large code changes.

src/test/csrc/verilator/emu.cpp
zephyrols: OK，I will restart a PR for this.

src/test/csrc/verilator/emu.cpp
zephyrols: Whether the database's parameters are closer to each other would be easier to maintain?

poemonsense:
Thanks merged

PR #600, opened 2025-03-20
Gsim

sashimi-yzh:


gsim.mk
```diff
@@ -0,0 +1,86 @@
...
+gsim-gen-cpp: $(GSIM_GEN_CSRC_DIR)/$(SIM_TOP)0.cpp
+
+##############################################
+### Building EMU from cpp model generated by GSIM
+##############################################
+
+GSIM_CONFIG_DIR = $(abspath ./config)
+VERILITOR_CSRC_DIR = $(abspath ./src/test/csrc/verilator)
+GSIM_OTHER_CSRC_DIR = $(abspath ./src/test/csrc/gsim)
+SIM_CXXFILES += $(shell find $(VERILITOR_CSRC_DIR) -name "*.cpp") \
```
poemonsense, line 43: I think this overrides the `SIM_CXXFILES` variable, which is shared by other simulators (VCS, Palladium, ...).

We need `GSIM_CXXFILES`, `GSIM_LDFLAGS`, ... here

src/test/csrc/verilator/emu.cpp
```diff
@@ -623,19 +636,37 @@ inline void Emulator::reset_ncycles(size_t cycles) {
...
+    dut_ptr->step();
+#endif // GSIM
   }
+
+#ifdef GSIM
+  // If timecmp starts from 0, a machine-level time interrupt will be issued
+  // immediately after the simulation, which makes MIP.mtip different from
+  // NEMU, and finally DiffTest reports an error. To fix this, just
+  // initialize timecmp with a value large enough.
+  dut_ptr->l_soc$socMisc$clint$timecmp_0 = 0x1234567887654321ul;
```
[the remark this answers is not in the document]
sashimi-yzh: The root cause is that NEMU should not assume `MIP.mtip == 0` after reset. If `timecmp == 0` after reset, then `MIP.mtip == 1`. Do you have any idea to fix this issue?

gsim.mk
```diff
@@ -0,0 +1,85 @@
+#***************************************************************************************
+# Copyright (c) 2024 Beijing Institute of Open Source Chip (BOSC)
```
poemonsense, line 2: Please help update the License header for years/institutions

src/main/scala/DPIC.scala
```diff
@@ -20,6 +20,7 @@ import chisel3.experimental.ExtModule
 import chisel3.reflect.DataMirror
 import chisel3.util._
 import difftest._
+import difftest.DifftestModule.createCppExtModule
```
poemonsense, line 23: @klin02 Do you consider merge these changes first? It looks difficult for us to review such a large PR

src/main/scala/common/LogPerfControl.scala
```diff
@@ -29,6 +29,22 @@ class LogPerfControl extends Bundle {
...
 
+  val cppExtModule =
+    """
+      |void LogPerfHelper (
+      |  uint64_t& timer,
+      |  uint8_t&  logEnable,
+      |  uint8_t&  clean,
+      |  uint8_t&  dump
+      |) {
+      |  timer     = difftest$$timer;
```
poemonsense, line 40: These are some gsim-specific variables. If this header is only used by gsim, it's OK.

src/test/csrc/verilator/emu.cpp
```diff
@@ -623,6 +636,14 @@ inline void Emulator::reset_ncycles(size_t cycles) {
 #ifdef COVERAGE_PORT_RESET
     dut_ptr->coverage_reset = dut_ptr->reset;
 #endif // COVERAGE_PORT_RESET
+#endif // VERILATOR
+
+#ifdef GSIM
```
poemonsense, line 641: Probably we need to implement this with C++ inheritence in the future. I'll work with @klin02 to refactor difftest in the near future. (Only comment here for future references)

poemonsense (commented):
CI fails because the formatting fails. Need to run `make format` before commit.

I'm writing some comments here for future references. Some of them need not to be fixed before merging.

src/test/csrc/verilator/emu.cpp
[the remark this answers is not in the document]
poemonsense: Fixed in https://github.com/OpenXiangShan/XiangShan/pull/4449

poemonsense:
@klin02 I think the current implementation in both Scala and C++ can be improved.

We need to provide better support for simulator-dependent or platform-dependent features in Scala. For example, when calling these helpers, we instantiate different models if they cannot be written in a compatible way for all simulators. Also, when the user calls difftest with GSIM, we should raise an explicit error if they are using Verilator in the next steps. Features like these require some parameterization or better code structure in the Scala.

Currently even in the Verilator C++ code, there're so many macros. We need to improve the code quality this year. Otherwise, the code will be very hard to maintain.

poemonsense:
Please let me know when this PR is expected to be merged.

We need to add tests after gsim is made public.

sashimi-yzh:
It is ready now. Refer to https://github.com/jaypiper/simulator/ for the latest version of gsim.

poemonsense (approved):

poemonsense:
Will add a test for gsim with rocket later

PR #601, opened 2025-03-20
perf(fpga): use spinLock instead of mutex and delet unwanted locks

xiaokamikami:
Since the overhead per operation is not large, mutexes are replaced by lighter locks. At the same time, by removing cv_wait from the function that controls group offset, the locks that are no longer needed are removed, which improves the processing bandwidth of mempool from  16161MB/S to more than 22450MB/S (The stress test was performed on the Intel(R) Core(TM) i7-11700 platform)


poemonsense (approved):

poemonsense:
I'm not sure how many people know all details here. Maybe we should later review all details again to fully understand it.

PR #602, opened 2025-03-27
feat(Uncache): add event and update goldenmem for some uncache cases

Maxpicca-Li:
## single core
store: PBMT>0 but PMA=main-mem, this case needs to update the golden memory.

## multi core

However, due to bus transmission delays, the moment when core0 receives the uncache (main-mem store) grant fire may be later than the moment when the data becomes visible to other cores. At this point, it is uncertain whether core1 has performed a load between these two moments. If such a load occurs, core1 in the DUT will read the updated data, whereas goldenmem will still hold the previous data, leading to a difftest mismatch.

Since the timing of data visibility to other cores varies across different simulators/protocols, the solution here is to set a `flag` in advance.

Specifically, after core0 commits the "store" but before the store becomes visible, such as when core0 issues an uncache (main-mem store) acquire request, the goldenmem flag is set. In this case, if there is a inconsistency in read results between NEMU, golden, and DUT, but the flag is set to 1, the DUT result should be considered correct, and both NEMU and goldenmem should be updated accordingly.

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -1011,7 +1011,7 @@ int Difftest::do_refill_check(int cacheid) {
             for (int j = 0; j < 8; j++) {
               Info("%016lx", dut_refill->data[j]);
             }
-             Info("\n");
+            Info("\n");
             update_goldenmem(dut_refill->addr, dut_refill->data, 0xffffffffffffffffUL, 64);
             proxy->ref_memcpy(dut_refill->addr, dut_refill->data, 64, DUT_TO_REF);
             return 0;
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -839,8 +839,7 @@ void Difftest::do_load_check(int i) {
             len = 8;
           }
         }
-        read_goldenmem(load_event.paddr, &golden, len);
-        read_goldenmem_flag(load_event.paddr, &golden_flag, len);
+        read_goldenmem_with_flag(load_event.paddr, &golden, &golden_flag, len);
         if (load_event.isLoad) {
           switch (len) {
             case 1:
@@ -871,7 +870,7 @@ void Difftest::do_load_check(int i) {
           Info("load check of uncache mm store flag\n");
           Info("  DUT data: 0x%lx, regWen: %d, refRegPtr: 0x%lx\n", commitData, regWen, refRegPtr);
           proxy->ref_memcpy(load_event.paddr, &commitData, len, DUT_TO_REF);
-          update_goldenmem(load_event.paddr, &commitData, mask, len, 0);
+          update_goldenmem(load_event.paddr, &commitData, mask, len);
           if (regWen) {
             *refRegPtr = commitData;
             proxy->sync(true);
@@ -975,8 +974,7 @@ int Difftest::do_refill_check(int cacheid) {
       return 0;
     }
     for (int i = 0; i < 8; i++) {
-      read_goldenmem(dut_refill->addr + i * 8, &buf, 8);
-      read_goldenmem_flag(dut_refill->addr + i * 8, &flag_buf, 8);
+      read_goldenmem_with_flag(dut_refill->addr + i * 8, &buf, &flag_buf, 8);
       if (dut_refill->data[i] != *((uint64_t *)buf)) {
 #ifdef CONFIG_DIFFTEST_CMOINVALEVENT
         if (cmo_inval_event_set.find(dut_refill->addr) != cmo_inval_event_set.end()) {
@@ -993,7 +991,7 @@ int Difftest::do_refill_check(int cacheid) {
             Info("%016lx", dut_refill->data[j]);
           }
           Info("\n");
-          update_goldenmem(dut_refill->addr, dut_refill->data, 0xffffffffffffffffUL, 64, 0);
+          update_goldenmem(dut_refill->addr, dut_refill->data, 0xffffffffffffffffUL, 64);
           proxy->ref_memcpy(dut_refill->addr, dut_refill->data, 64, DUT_TO_REF);
           cmo_inval_event_set.erase(dut_refill->addr);
           return 0;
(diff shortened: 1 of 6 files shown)


poemonsense (approved):

Anzooooo (changes requested):
I still have questions about nc's difftest solution, I'll discuss it with YQ before deciding.

poemonsense (dismissed):

src/test/csrc/difftest/goldenmem.h
```diff
@@ -28,20 +29,23 @@ typedef uint64_t uint64_t;
...
 extern uint8_t *pmem;
+extern uint8_t *pmem_flag;
 
 void init_goldenmem();
 void goldenmem_finish();
 
-extern "C" void update_goldenmem(uint64_t addr, void *data, uint64_t mask, int len);
+extern "C" void update_goldenmem(uint64_t addr, void *data, uint64_t mask, int len, uint8_t flag);
 extern "C" void read_goldenmem(uint64_t addr, void *data, uint64_t len);
+extern "C" void read_goldenmem_flag(uint64_t addr, void *flag, uint64_t len);
```
forever043, line 39: I `DO-NOT` prefer this API change.
KEEP core APIs `stable` was essential for a mature project (a.k.a. our difftest).

For the majority callers of `update_goldenmem()`, they use the `flag == 0` args. And the only user of this new `flag` parameter was `Difftest::do_golden_memory_update`. This API change will cause lot of use-less(but required) modifies to all other callers.

I suggest to add an extra API for this purpose (e.g. `update_goldenmem_with_flags`, or just `__update_goldenmem(..., flag), and give default `flag = 0` for the old API as follow:
```
void __update_goldenmem(uint64_t addr, void *data, uint64_t mask, uint6_t flag) {
   // original code from update_goldenmem()
}
void update_goldenmem(uint64_t addr, void *data, uint64_t mask) {
    __update_goldenmem(addr, data, mask, 0 /* default */);
}
```

src/test/csrc/difftest/difftest.cpp
```diff
@@ -838,11 +840,24 @@ void Difftest::do_load_check(int i) {
           }
         }
         read_goldenmem(load_event.paddr, &golden, len);
+        read_goldenmem_flag(load_event.paddr, &golden_flag, len);
```
forever043, line 843: As we always call `read_goldenmem_flag()` after `read_goldenmem()` at most time, we can merge the impl of these 2 API into a single one, but keep 2* API variant (for the same reason of KEEP ABI stable):

```
void read_goldenmem_flag(uint64_t addr, void *data, void *flag, uint64_t len) {
    // original read_goldenmem() code
    if (flag != NULL) {
        *flag = xxx;
    }
}

void read_goldenmem(uint64_t addr, void *data, uint64_t len) {
    read_goldenmem_flag(addr, data, NULL, len);
}
```

Then, we can merge all `seq call to read_goldenmem/read_goldenmem_flag`  with single call to `read_goldenmem_flag`

src/test/csrc/difftest/difftest.cpp
```diff
@@ -851,6 +866,16 @@ void Difftest::do_load_check(int i) {
             *refRegPtr = commitData;
             proxy->sync(true);
           }
+        } else if (load_event.isLoad && golden_flag != 0) {
+          // goldenmem check failed, but the flag is set, so use DUT data to reset
+          Info("load check of uncache mm store flag\n");
+          Info("  DUT data: 0x%lx, regWen: %d, refRegPtr: 0x%lx\n", commitData, regWen, refRegPtr);
```
forever043, line 872: @poemonsense  do we have some DEBUG level output API (e.g. `Debug("")`).
I'm a little worried that these logs will happen in large numbers, making the difftest log difficult to read.

src/test/csrc/difftest/goldenmem.h
```diff
@@ -28,20 +29,23 @@ typedef uint64_t uint64_t;
 typedef uint64_t word_t;
 
 extern uint8_t *pmem;
+extern uint8_t *pmem_flag;
 
 void init_goldenmem();
 void goldenmem_finish();
 
-extern "C" void update_goldenmem(uint64_t addr, void *data, uint64_t mask, int len);
+extern "C" void update_goldenmem(uint64_t addr, void *data, uint64_t mask, int len, uint8_t flag);
```
forever043, line 37: Will we add other `flag` in the future? If not sure, I suggest to use a more explicit name (e.g. `skip_load_check`) to express the purpose of this flag.

src/test/csrc/difftest/difftest.cpp
poemonsense: For now, No.

As discussed in https://github.com/OpenXiangShan/difftest/pull/546#pullrequestreview-2539545197, we should implement this feature in the future. But I think recently (at least this month), it seems both Kunlin and I don't have much time for it.

src/test/csrc/difftest/goldenmem.h
```diff
@@ -28,20 +29,23 @@ typedef uint64_t uint64_t;
...
 extern uint8_t *pmem;
+extern uint8_t *pmem_flag;
 
 void init_goldenmem();
 void goldenmem_finish();
 
-extern "C" void update_goldenmem(uint64_t addr, void *data, uint64_t mask, int len);
+extern "C" void update_goldenmem(uint64_t addr, void *data, uint64_t mask, int len, uint8_t flag = 0);
 extern "C" void read_goldenmem(uint64_t addr, void *data, uint64_t len);
+extern "C" void read_goldenmem_with_flag(uint64_t addr, void *data, void *flag, uint64_t len);
```
forever043, line 39: the c++ default args was a good idea here~
maybe we can use same mechanism for `read_goldenmem()`  by moving `flags` to last and give `NULL` as default?

forever043: this is not mandatory, and I just prefer to implement these function-related interfaces in the same way

Anzooooo (approved):

Anzooooo (approved):

forever043 (approved):
LGTM

PR #611, opened 2025-04-25
fix(difftest): fix alignment for C++ type declarations

NewPaulWalker:
When generating `arrayType`, not all 8-bit-aligned types are valid in C++.

For example, if someone defines a 41-bit-wide signal, it may result in generating an unsupported C++ type like uint48_t instead of a valid uint64_t.

This change ensures that for any given bit width, a C++-supported type is generated.

### DIFF
--- a/src/main/scala/Bundles.scala
+++ b/src/main/scala/Bundles.scala
@@ -48,8 +48,8 @@ sealed trait DifftestBaseBundle extends Bundle {
   }
 }
 
-class DeltaElem(elemBytes: Int) extends DifftestBaseBundle {
-  val data = UInt((elemBytes * 8).W)
+class DeltaElem(elemWidth: Int) extends DifftestBaseBundle {
+  val data = UInt(elemWidth.W)
 }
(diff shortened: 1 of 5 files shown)


klin02:
We have used the same alignment in many other places. If such a situation (like a 41-bit-wide signal) does exist, we will make the changes together in this PR.

NewPaulWalker:
> We have used the same alignment in many other places. If such a situation (like a 41-bit-wide signal) does exist, we will make the changes together in this PR.

It's OK! Should I close this PR directly now, or temporarily convert it to a draft?

poemonsense:
> > How is this PR going?
> 
> I think if necessary, we can replace all `(size + 7) / 8 * 8` with a common method, or simply leave it as it is.

We need a conclusion for each PR whose author has done his/her work.

Either we, as maintainers, will accept it in the future or now (then do it ASAP) or not accept it (and close it)

poemonsense (approved):

PR #612, opened 2025-04-25
chore: place init_flash after init_core

NewPaulWalker:
First, this does not introduce any functional changes.

Second, the reason for this modification is that with the introduction of https://github.com/OpenXiangShan/XiangShan/pull/4626, the simulation output includes some less elegant messages:

> Using simulated 32768B flash
> Core  0's Commit SHA is: 6683fc49ec, dirty: 1
> Using simulated 8386560MB RAM

This change makes the output look better.

### DIFF
--- a/src/test/csrc/verilator/emu.cpp
+++ b/src/test/csrc/verilator/emu.cpp
@@ -377,8 +377,6 @@ Emulator::Emulator(int argc, const char *argv[])
   if (enable_simjtag) {
     jtag_init();
   }
-  // init flash
-  init_flash(args.flash_bin);
 
 #if VM_TRACE == 1
   if (args.enable_waveform) {
@@ -400,6 +398,9 @@ Emulator::Emulator(int argc, const char *argv[])
 
   // init core
   reset_ncycles(args.reset_cycles);
+
+  // init flash
+  init_flash(args.flash_bin);
 
   // init ram
   uint64_t ram_size = DEFAULT_EMU_RAM_SIZE;


src/test/csrc/verilator/emu.cpp
poemonsense, line 1: If you git blame this line, you will see this: https://github.com/OpenXiangShan/difftest/commit/c19a798b7ce9a8e8beff7c1d10fe10dcd0fb7aef

But I don't remember why. It seems at that time we didn't write down the reasons.

src/test/csrc/verilator/emu.cpp
```diff
@@ -377,8 +377,6 @@ Emulator::Emulator(int argc, const char *argv[])
   if (enable_simjtag) {
     jtag_init();
   }
-  // init flash
```
poemonsense, line 380: If you git blame this line, you will see this: https://github.com/OpenXiangShan/difftest/commit/c19a798b7ce9a8e8beff7c1d10fe10dcd0fb7aef

But I don't remember why. It seems at that time we didn't write down the reasons.

NewPaulWalker: Thanks for the detailed review. I'll run CI tests on XiangShan with this commit to check if the previous issue can be reproduced. If not, we'll prioritize ensuring functional correctness for now.

poemonsense: Thanks

NewPaulWalker: Unfortunately, it seems that we weren't able to reproduce the bug https://github.com/OpenXiangShan/XiangShan/actions/runs/14705130674. I think we'll have to keep the previous changes and close this PR for now?

poemonsense: I think both are fine. Since it cannot be reproduced and we don't remember the root cause, we can merge it first. If in the future the bug occurs, we then fix it again.

Also, if you don't mind the order of printf, I prefer not changing the existing code

NewPaulWalker: I don't really mind the order of the `printf`; either way works for me. I'll go ahead and close this PR. Thanks for your help!

PR #615, opened 2025-04-27
feat(build): Prepare for XiangShan docker support

forever043:


verilator.mk
```diff
@@ -136,7 +139,9 @@ endif
 	@mkdir -p $(@D)
 	@echo -e "\n[verilator] Generating C++ files..." >> $(TIMELOG)
 	@date -R | tee -a $(TIMELOG)
-	$(TIME_CMD) verilator $(VERILATOR_FLAGS) -Mdir $(@D) $^ $(EMU_DEPS)
+	$(TIME_CMD) $(VERILATOR) $(VERILATOR_FLAGS) --Mdir $(@D) $^ $(EMU_DEPS)
+	@sed -i -e 's/$(subst /,\/,$(NOOP_HOME))/$$(NOOP_HOME)/g' \
```
poemonsense, line 143: What is this

forever043: 1. fix wrong `-Mdir` to `--Mdir` (although it work now)
2. use VERILATOR variable instead of verilator to support override outside
3. change abs-path in generated VSimTop.mk to `$(NOOP_HOME)/xxx`, then we can move the build directory and reset NOOP_HOME env

all these changes are pre-requirements of docker build support [OpenXiangShan/PR#4538](https://github.com/OpenXiangShan/XiangShan/pull/4538)

poemonsense: > 3. change abs-path in generated VSimTop.mk to `$(NOOP_HOME)/xxx`, then we can move the build directory to other place by reset NOOP_HOME env

I see. But this would require that NOOP_HOME is always specified, even after verilator exits. I'm not sure if this is OK?

forever043: > > 3. change abs-path in generated VSimTop.mk to `$(NOOP_HOME)/xxx`, then we can move the build directory to other place by reset NOOP_HOME env
> 
> I see. But this would require that NOOP_HOME is always specified, even after verilator exits. I'm not sure if this is OK?

NOOP_HOME was not required, the abs-path was set as default in VSimTop.mk
e.g.:
```
NOOP_HOME ?= /xxx/yyy
default: $(NOOP_HOME)/emu
...
```

poemonsense: Great

poemonsense (approved):

PR #616, opened 2025-04-28
fix(VCS): fix VCS compile std feature

cz4e:
* use c++20 for SIMV simulation


Anzooooo:
The pr is closed for compatibility reasons.
We will modify the code to make it compatible with the c++ 11 specification.

PR #618, opened 2025-04-29
FPGA: fix arrangement and parsing of Batch packet

xiaokamikami:
It supports the batch data packets running on the FPGA-diff platform, and adds the stack print when the thread segment error occurs, as well as the fpga performance hint print, and adds the flash through the parameter


src/test/csrc/fpga/fpga_main.cpp
```diff
@@ -156,14 +179,23 @@ void cpu_endtime_check() {
         }
       }
     }
+  } else {
+    auto trap = difftest[0]->get_trap_event();
+    cycleCnt = trap->cycleCnt;
+  }
+
+  if (cycleCnt % 0xFFFFFF) {
```
poemonsense, line 187: should we preserve these debug print in the open-source version?

src/test/csrc/fpga/xdma.cpp
```diff
@@ -138,23 +157,30 @@ void FpgaXdma::stop_thansmit_thread() {
 }
 
 void FpgaXdma::read_xdma_thread(int channel) {
+  const uint64_t xdma_get_len = ((sizeof(FpgaPackgeHead) + 128) / 128) * 128; // send width to be calculated by mod up
```
poemonsense, line 160: If round up, should we use (X+127)/128?

xiaokamikami: We're aligning memory lengths, not arrays, so we shouldn't need to decrement by 1

src/test/csrc/fpga/fpga_main.cpp
xiaokamikami: I printed the speed information at the end

src/test/csrc/fpga/fpga_main.cpp
```diff
@@ -138,18 +145,36 @@ void fpga_step() {
...
+  perf_run_sec_start = ts.tv_sec;
+}
+
+void fpga_perfcnt_print(uint64_t cycleCnt) {
+  uint64_t perf_run_sec;
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  perf_run_sec = ts.tv_sec - perf_run_sec_start;
+  float speed = cycleCnt / 1000 / (float)perf_run_sec;
+  printf("\rFpga speed: %.1f K Cycle", speed);
```
klin02, line 160: We have difftest_perf_init/finish in csrc/common/perf.cpp, no need to declare fpga_perf_init/finish here.

src/test/csrc/fpga/fpga_main.cpp
```diff
@@ -138,18 +145,36 @@ void fpga_step() {
...
 void cpu_endtime_check() {
+  uint64_t cycleCnt = 0;
   if (max_instrs != 0) { // 0 for no limit
     for (int i = 0; i < NUM_CORES; i++) {
       if (core_end_info.core_trap[i])
         continue;
       auto trap = difftest[i]->get_trap_event();
-      if (max_instrs < trap->instrCnt) {
+      cycleCnt = trap->cycleCnt;
+      if (max_instrs < cycleCnt) {
```
klin02, line 171: Why compare Instructions and Cycles here? Maybe you want to compare max_instrs and trap->instrCnt.

src/test/csrc/fpga/xdma.cpp
klin02: I still wonder why we add 128 here. Seems we need only BatchSize + 1 Byte + padding (if any), instead of 128 Byte here.

I think we can simply round up in xdma.h. For example, we can declare as 
```
#define DMA_PADDING ... // calculate here, note the value may be 0
typedef struct FpgaPackgeHead {
  uint8_t idx;
  uint8_t packge[DMA_DIFF_PACKGE_LEN];
#if `DMA_PADDING != 0
  uint8_t padding[DMA_PADDING ];
#endif
} FpgaPackgeHead;
```

Then, we can simply read to it in line 167, no need to memcpy twice in line 167 and 176.

src/test/csrc/fpga/xdma.cpp
```diff
@@ -138,23 +157,30 @@ void FpgaXdma::stop_thansmit_thread() {
...
 #else
 #ifdef CONFIG_DIFFTEST_BATCH
+    memcpy(packge, xdma_packge, sizeof(FpgaPackgeHead));
     v_difftest_Batch(packge->diff_packge);
 #elif defined(CONFIG_DIFFTEST_SQUASH)
     //TODO: need automatically generates squash data parsing implementations
 #endif // CONFIG_DIFFTEST_BATCH
 #endif // USE_THREAD_MEMPOOL
   }
+  free(xdma_packge);
```
klin02, line 183: We should not malloc and free a large space every time xdma_read, which will consumes more time.
Maybe we can add the malloc in some `init` function and free them at the end of simulation.

src/test/csrc/fpga/xdma.cpp
```diff
@@ -138,23 +157,30 @@ void FpgaXdma::stop_thansmit_thread() {
 }
 
 void FpgaXdma::read_xdma_thread(int channel) {
+  const uint64_t xdma_get_len = ((sizeof(FpgaPackgeHead) + 128) / 128) * 128; // send width to be calculated by mod up
+  uint8_t *xdma_packge = (uint8_t *)malloc(xdma_get_len);
   FpgaPackgeHead *packge = (FpgaPackgeHead *)malloc(sizeof(FpgaPackgeHead));
+  int debug_count = 0;
```
klin02, line 163: remove the debug info.

src/test/csrc/fpga/xdma.cpp
```diff
@@ -138,23 +157,30 @@ void FpgaXdma::stop_thansmit_thread() {
...
 #ifdef USE_THREAD_MEMPOOL
-    if (xdma_mempool.write_free_chunk(idx, (char *)&packge) == false) {
+    uint8_t idx = xdma_packge[0];
+    if (xdma_mempool.write_free_chunk(idx, (char *)&xdma_packge) == false) {
       printf("It should not be the case that no available block can be found\n");
       assert(0);
     }
 #else
 #ifdef CONFIG_DIFFTEST_BATCH
+    memcpy(packge, xdma_packge, sizeof(FpgaPackgeHead));
```
klin02, line 176: xdma_package[0] is idx in line 169, we should not view it as Batch Package directly.

xiaokamikami: idx need to be checked at readout time in case the packet order is out of order

src/test/csrc/fpga/xdma.cpp
xiaokamikami: xdma_read goes into a loop and does not request memory repeatedly

src/test/csrc/fpga/fpga_main.cpp
```diff
@@ -139,6 +143,7 @@ void fpga_step() {
 }
 
 void cpu_endtime_check() {
+  uint64_t cycleCnt = 0;
```
klin02, line 146: Unused value.

src/test/csrc/fpga/xdma.cpp
klin02: Unused Value.

src/test/csrc/fpga/xdma.h
```diff
@@ -29,13 +29,19 @@
 #include <vector>
 
 #ifdef CONFIG_DIFFTEST_BATCH
-#define DMA_DIFF_PACKGE_LEN (CONFIG_DIFFTEST_BATCH_BYTELEN + 1)
+#define DMA_DIFF_PACKGE_LEN (CONFIG_DIFFTEST_BATCH_BYTELEN)
 #elif defined(CONFIG_DIFFTEST_SQUASH)
 #define DMA_DIFF_PACKGE_LEN 1280 // XDMA Min size
 #endif
+#define DMA_PADDING \
+  (((DMA_DIFF_PACKGE_LEN + 1 + 63) / 64) * 64 - (DMA_DIFF_PACKGE_LEN - 1)) // send width to be calculated by mod up
```
klin02, line 37: Maybe `- (DMA_DIFF_PACKGE_LEN + 1)` ? We can also add annotation to explain the `1` means the idx of package.

klin02 (approved):

PR #621, opened 2025-05-12
refactor(difftest): reuse common nstep logic for emu/simv/fpga

klin02:
Previously, we use various difftest/simv/fpga_nstep functions for Verilator/VCS/FPGA, resulting in duplicated code and inconsistent simulation conditions.

For example, in the multi-core scenario, VCS/FPGA warmedup and exceed at the same number of instructions for each core, while Verilator warmup and exceed when a certain core reaching instrs.

This change reuse the common difftest_nstep as the comparison logic and aligned the simulation conditions of VCS/FPGA with those of Verilator.

We also the issues of VCS where SIMV_WARMUP did not return correctly to DUT and where the number of instructions was calculated incorrectly during soft warmup.


xiaokamikami:
The changes now make the code look cleaner. Good commit

xiaokamikami (approved):

poemonsense (approved):

PR #624, opened 2025-05-31
FPGA: support simulate fpga-host with dut

klin02:
This change supports software simulation of FPGA-host interacting with DUT in a xdma-like way. We use a simv process to simulate DUT and send difftest_data with xdma-like pcie. The simv process will interacting with fpga-host by share memory acting as xdma device.

Code of the xdma_ctrl.v is modified from the code of fengkehan <fengkehan@bosc.ac.cn> and fix some bugs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -331,11 +331,20 @@ jobs:
             cd NutShell && rm -rf difftest && cp -r $GITHUB_WORKSPACE .
             echo "NOOP_HOME=$(pwd)" >> $GITHUB_ENV
 
-      - name: FPGA-difftest Build
+      - name: FPGA-difftest Build (no THREAD, no ASYNC_CLK)
         run: |
             cd $NOOP_HOME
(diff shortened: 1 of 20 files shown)


klin02:
Note when THREAD_MEMPOOL is enabled, the fpga-host will only start difftest when num of recieved data package greater than 256. However, the microbench workload only produce less than 256 package during running, and slowly produce other package after exiting, which resulting in a longer time in CI job.

poemonsense (approved):

xiaokamikami (approved):

xiaokamikami:
> Note when THREAD_MEMPOOL is enabled, the fpga-host will only start difftest when num of recieved data package greater than 256. However, the microbench workload only produce less than 256 package during running, and slowly produce other package after exiting, which resulting in a longer time in CI job.

Yes, this is because the current idx limit of the sent packet is 256, we use 256 to divide the whole memory block, and the divided data can achieve lock-free operation

PR #633, opened 2025-06-25
fix(mem): support both async and sync read

klin02:
GSIM only support async read of DiffTestMEM to ensure combinational logic inside extmodule, while Palladium only support sync read to only trigger DPI-C in clock posedge.

This change support both async and sync read, which can be switched by port a_sync with value 1/0 when defining GSIM or not.

### DIFF
--- a/src/main/scala/common/Mem.scala
+++ b/src/main/scala/common/Mem.scala
@@ -68,6 +68,7 @@ private trait HasReadPort { this: ExtModule =>
     val enable = Input(Bool())
     val index = Input(UInt(64.W))
     val data = Output(UInt(64.W))
+    val async = Output(Bool())
   })
 
   val r_dpic =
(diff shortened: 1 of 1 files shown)


src/main/scala/common/Mem.scala
```diff
@@ -139,15 +155,13 @@ private trait HasWritePort { this: ExtModule =>
 
   val w_func =
     """
+      |always @(posedge clock) begin
       |`ifndef DISABLE_DIFFTEST_RAM_DPIC
-      |if (w_enable) begin
-      |  difftest_ram_write(w_index, w_data, w_mask);
```
poemonsense, line 144: no need to change this?

Using begin and end is good for tracking changes

poemonsense (approved):

src/main/scala/common/Mem.scala
klin02: OK. I will fix that.

poemonsense (approved):

PR #639, opened 2025-07-11
common: add the create_noop_filename function

poemonsense:
This function can be used to generate formatted filename under NOOP.


poemonsense:
A minor change before fixing waveform

src/test/csrc/common/common.cpp
```diff
@@ -125,3 +126,36 @@ extern "C" void enable_sim_verbose() {
...
+    }
+#endif
+    assert(noop_home != NULL);
+
+    noop_len = snprintf(noop_filename, sizeof(noop_filename), "%s/build/", noop_home);
+  }
+
+  // get formatted time
+  time_t now = time(NULL);
+  int time_len = strftime(noop_filename + noop_len, sizeof(noop_filename) - noop_len, "%F-%H-%M-%S", localtime(&now));
```
poemonsense, line 151: Note, the filename format is changed to avoid `:` and `@` in file names.

PR #642, opened 2025-07-16
st_tool: add config on enable fpga-diff for interface

xiaokamikami:


forever043 (approved):
a hack solution, but it can temporarily resolve our problem

LGTM

PR #647, opened 2025-07-18
fix(logdb_filename): remove undeclared arguments

klin02:


klin02:
See declarations in https://github.com/OpenXiangShan/difftest/blob/eb10cb0973be6a39ca40bce69c87c2bf4fc92fc3/src/test/csrc/verilator/emu.h#L118

PR #648, opened 2025-07-20
emu: add an abstraction class for RTL simulators

poemonsense:
This commit adds a standard abstraction for different RTL simulators for emu-mode, where the simulation is driven by C++ code at the top.

For various RTL simulators, they only need to implement the mandatory and some optional methods to work on DiffTest.


klin02 (dismissed):

poemonsense:
Already tested with GSIM. The compilation can finish correctly once GSIM fixes some of its bugs. However, the simulation results seem incorrect now.

klin02 (approved):

PR #650, opened 2025-07-22
Preprocess: split commitData and vecCommitData

klin02:


klin02:
It helps reduce area of DiffTest when removing vecWriteBack check.

PR #652, opened 2025-07-24
fix(gsim): add missing r_async for cppExtModule

klin02:


src/main/scala/common/Mem.scala
```diff
@@ -122,9 +122,14 @@ private trait HasReadPort { this: ExtModule =>
     """
       |uint8_t   r_enable,
       |uint64_t  r_index,
-      |uint64_t& r_data""".stripMargin
+      |uint64_t& r_data
```
poemonsense: should it be `r_data,`

klin02: My mistake:) Will fix soon

poemonsense (approved):

PR #655, opened 2025-08-05
fpga: optimize timing and area with BRAM

xiaokamikami:
The timing of sending module buffer by axis is optimized by dismantling the register buffer


xiaokamikami:
After testing, the new method has no impact on speed. The LUT used by the sending module has decreased from 17w to 2w, the BRAM usage has increased by 448 (11% of the total resources of VU19P), and the timing pressure of the wiring has decreased. Therefore, the BRAM solution can be considered.

klin02:
After tested in FPGA, this PR can be merged.

src/test/vsrc/fpga_sim/dual_buffer_bram.v
```diff
@@ -0,0 +1,55 @@
...
+  // TODO: Implement automatic BRAM stitching when more width configurations are available
+
+  localparam BLOCK_RAM_DATA_WIDTH = 4000;
+
+  wire [BLOCK_RAM_DATA_WIDTH - 1:0] rd_data_split [3:0];
+  wire [BLOCK_RAM_DATA_WIDTH - 1:0] wr_data_split [3:0];
+  assign wr_data_split[0] = wr_data[ 3999:    0];
+  assign wr_data_split[1] = wr_data[ 7999: 4000];
+  assign wr_data_split[2] = wr_data[11999: 8000];
+  assign wr_data_split[3] = wr_data[15999:12000];
```
klin02: Here, the bit width has been hardcoded, so the DATA_WIDTH parameter doesn't work, and passing different parameters will lead to inconsistency. Try using the "generate" statement instead?

src/test/vsrc/fpga_sim/xdma_ctrl.v
```diff
@@ -50,20 +49,51 @@ import "DPI-C" function bit v_xdma_tready();
...
     assign axi_tdata = reg_axi_tdata;
     assign axi_tvalid = reg_axi_tvalid;
     assign axi_tkeep = 64'hffffffff_ffffffff;
     assign axi_tlast = reg_axi_tlast;
 
     wire both_full = &buffer_valid;
+    dual_buffer_bram #(
+      .DATA_WIDTH(DATA_WIDTH),
+      .NUM_PACKETS_PER_BUFFER(NUM_PACKETS_PER_BUFFER),
+      .ADDR_WIDTH(3)
```
klin02: The ADDR_WIDTH here seems unconfigured. Use params like `NUM_PACKETS_PER_BUFFER`.
Also when we repeatedly instantiate the same module, using `generate` helps avoid redundant code.

src/test/vsrc/fpga_sim/xdma_ctrl.v
```diff
@@ -163,13 +192,18 @@ import "DPI-C" function bit v_xdma_tready();
...
                     end else if (!last_pkt & last_send) begin
-                        mix_data <= {dual_buffer[rd_buf][rd_pkt_cnt], 8'b0};
+                        mix_data <= {dual_buffer_rd_data_mux, 8'b0};
                         rd_pkt_cnt <= rd_pkt_cnt + 1'b1;
                         datalen <= 0;
                     end else begin
                         datalen <= datalen + 1'b1;
                         mix_data <= mix_data >> AXIS_DATA_WIDTH;
                     end
+                end else if (~reg_axi_tvalid)begin
```
klin02, line 195: Seems strange here? The `reg_axi_tvalid` register is reset to 0 only when the state is "Done".

xiaokamikami: When entering TRANSFER for the first time, reg_axi_tvalid = 0

PR #661, opened 2025-08-22
fix(wiring): teat chisel version in a more concrate way

Tang-Haojin:
`BoringUtils.tapAndRead` is introduced in chisel 6. Previous testing method ignored chisel 5.


Tang-Haojin:
I think it’s fine to drop Chisel 3 support. XiangShan removed it a year ago, so this change won’t affect us. If we decide to go ahead, I can open a PR.

klin02 (approved):

PR #675, opened 2025-09-12
fpga: add required fpga rtl and seperate from fpga-sim

klin02:


poemonsense (approved):
Need converting to Chisel in the future

xiaokamikami (approved):

PR #678, opened 2025-09-12
Batch: group multi-bundle to 8 and cat single-bundle

klin02:
BatchCluster collected same-type bundles with O(n^2) complexity, resulting a significant area when grouping a lot of same-type bundles. For example, there is 48 StoreEvents in Default Diff, and 64 regfile elems with Delta enabled.

This patch reduces complexity by grouping multi-bundle into groups of 8, such as 6x8 cluster for StoreEvent. And Batch also directly concat single-bundle instead of using extra shifter.

Overall, this change reduces 1/3 area for BatchCluster when BasicDiff.


poemonsense (approved):

poemonsense (commented):
The code is hard to read. Need improve its quality in terms of PPA and readability

PR #681, opened 2025-09-16
feat(simfrontend): add support for the simfrontend(ideal)

Anzooooo:
Currently, I have implemented an ideal decoupled frontend for Xiangshan to prepare for performance debugging of the backend and memory access.

These modifications have been verified to work on my local machine, and they have achieved the desired results.

However, I'm still unsure about how to organize my code.
I'm uncertain whether it should be placed directly in the difftest repository. Therefore, this PR also serves as a request for advice: Does Mr. Xu @poemonsense have any suggestions on where to place this code?

---

This commit(https://github.com/OpenXiangShan/difftest/commit/a55a474e5dce06010190bc52d619251ca64c174e) enables normal generation of skip instruction traces in NEMU, regardless of whether changes are made, and will not affect the normal operation of difftest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,6 +239,14 @@ SIM_CXXFLAGS += -I$(abspath $(PLUGIN_CSRC_DIR)/runahead)
 SIM_CXXFILES += $(shell find $(PLUGIN_CSRC_DIR)/runahead -name "*.cpp")
 endif
 
+# SimFrontend plugin
+ifeq ($(ENABLE_SIMFRONTEND), 1)
+TRACE_CSR_DIR = $(abspath ./src/test/csrc/plugin/simfrontend)
+SIM_CXXFILES += $(shell find $(TRACE_CSR_DIR) -name "*.cpp")
+SIM_CXXFLAGS += -I$(TRACE_CSR_DIR) -DPLUGIN_SIMFRONTEND
+endif
+
+
 # Check if XFUZZ is set
 ifeq ($(XFUZZ), 1)
 XFUZZ_HOME_VAR = XFUZZ_HOME
--- a/src/test/csrc/verilator/emu.cpp
+++ b/src/test/csrc/verilator/emu.cpp
@@ -39,6 +39,9 @@
 #ifdef ENABLE_IPC
 #include <sys/stat.h>
 #endif
+#ifdef PLUGIN_SIMFRONTEND
+#include "simfrontend.h"
+#endif // PLUGIN_SIMFRONTEND
 
 extern remote_bitbang_t *jtag;
 
@@ -87,6 +90,9 @@ static inline void print_help(const char *file) {
   printf("  -X, --fork-interval=NUM    LightSSS snapshot interval (in seconds), default: 10\n");
   printf("      --overwrite-nbytes=N   set valid bytes, but less than 0xf00, default: 0xe00\n");
   printf("      --overwrite-auto       overwrite size is automatically set of the new gcpt\n");
+#ifdef PLUGIN_SIMFRONTEND
+  printf("      --instr-trace          Setting the trace of instructions for SimFrontEnd\n");
+#endif // PLUGIN_SIMFRONTEND
   printf("      --force-dump-result    force dump performance counter result in the end\n");
   printf("      --load-snapshot=PATH   load snapshot from PATH\n");
   printf("      --enable-snapshot      enable simulation snapshots\n");
@@ -161,6 +167,7 @@ inline EmuArgs parse_args(int argc, const char *argv[]) {
     { "dramsim3-ini",      1, NULL,  0  },
     { "dramsim3-outdir",   1, NULL,  0  },
     { "overwrite-auto",    1, NULL,  0  },
+    { "instr-trace",       1, NULL,  0  },
     { "seed",              1, NULL, 's' },
     { "max-cycles",        1, NULL, 'C' },
     { "fork-interval",     1, NULL, 'X' },
@@ -268,6 +275,7 @@ inline EmuArgs parse_args(int argc, const char *argv[]) {
             break;
 #endif
           case 27: args.overwrite_nbytes_autoset = true; continue;
+          case 28: args.instr_trace = optarg; continue;
         }
         // fall through
       default: print_help(argv[0]); exit(0);
@@ -388,6 +396,10 @@ Emulator::Emulator(int argc, const char *argv[])
       dut_ptr->waveform_init(waveform_clock);
     }
   }
+
+#ifdef PLUGIN_SIMFRONTEND
+  init_sim_frontend(args.instr_trace);
+#endif // PLUGIN_SIMFRONTEND
 
   // init core
   reset_ncycles(args.reset_cycles);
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -157,6 +157,14 @@ SIM_CXXFILES += $(BUILD_DIR)/constantin.cpp
 SIM_CXXFLAGS += -I$(BUILD_DIR) -DENABLE_CONSTANTIN
 endif
 
+# SimFrontend
+ENABLE_SIMFRONTEND ?= 1
+ifeq ($(ENABLE_SIMFRONTEND), 1)
+TRACE_CSR_DIR = $(abspath ./src/test/csrc/simfrontend)
```
poemonsense: Maybe `src/test/csrc/plugin/simfrontend` is better

Makefile
```diff
@@ -157,6 +157,14 @@ SIM_CXXFILES += $(BUILD_DIR)/constantin.cpp
...
+SIM_CXXFILES += $(shell find $(TRACE_CSR_DIR) -name "*.cpp")
+SIM_CXXFLAGS += -I$(TRACE_CSR_DIR) -DENABLE_SIMFRONTEND
```
poemonsense: `PLUGIN_SIMFRONTEND`

src/test/csrc/simfrontend/debug.h
```diff
@@ -0,0 +1,29 @@
+#ifndef TRACE_DEBUG_H
```
poemonsense: Add license headers

Makefile
poemonsense: I'll change other plugins to `PLUGIN_*` macro in the future. So please use this macro

src/test/csrc/difftest/refproxy.h
```diff
@@ -221,8 +221,8 @@ class RefProxy : public AbstractRefProxy {
     return regs_vec.value + src;
   }
 #endif // CONFIG_DIFFTEST_ARCHVECREGSTATE
-  inline void sync(bool is_from_dut = false) {
-    ref_regcpy(&regs_int, is_from_dut, is_from_dut);
+  inline void sync(bool is_from_dut = false, bool is_skip = false) {
```
poemonsense: This `regcpy` function has one more argument in Spike (https://github.com/OpenXiangShan/riscv-isa-sim/blob/difftest/difftest/difftest.cc#L573, https://github.com/OpenXiangShan/riscv-isa-sim/commit/78399985c5dd9c3bf93f34354f4f7699d975c013):

`void difftest_regcpy(diff_context_t* dut, bool direction, bool on_demand)` When `on_demand` is set to true, only registers that differ from Spike are copied. I don't really remember why I added this argument.

Is the `on_demand` the same as `is_skip` here?

If not, we may need to align this function first, before merging this simfrontend.

src/test/csrc/verilator/emu.cpp
```diff
@@ -39,6 +39,9 @@
 #ifdef ENABLE_IPC
 #include <sys/stat.h>
 #endif
+#ifdef ENABLE_SIMFRONTEND
```
poemonsense: `PLUGIN_SIMFRONTEND`

src/test/csrc/verilator/emu.cpp
```diff
@@ -87,6 +90,7 @@ static inline void print_help(const char *file) {
   printf("  -X, --fork-interval=NUM    LightSSS snapshot interval (in seconds), default: 10\n");
   printf("      --overwrite-nbytes=N   set valid bytes, but less than 0xf00, default: 0xe00\n");
   printf("      --overwrite-auto       overwrite size is automatically set of the new gcpt\n");
+  printf("      --instr-trace          Setting the trace of instructions for SimFrontEnd\n");
```
poemonsense, line 94: `#ifdef PLUGIN_SIMFRONTEND`

src/test/csrc/verilator/emu.cpp
```diff
@@ -268,6 +273,7 @@ inline EmuArgs parse_args(int argc, const char *argv[]) {
             break;
 #endif
           case 27: args.overwrite_nbytes_autoset = true; continue;
+          case 28: args.instr_trace = optarg; continue;
```
poemonsense, line 278: `#ifdef PLUGIN_SIMFRONTEND`

verilator.mk
```diff
@@ -123,6 +123,7 @@ EMU_MK  = $(EMU_DIR)/V$(EMU_TOP).mk
 EMU_DEPS  := $(SIM_VSRC) $(EMU_CXXFILES)
 EMU_HEADERS := $(shell find $(EMU_CSRC_DIR) -name "*.h")     \
                $(shell find $(SIM_CSRC_DIR) -name "*.h")     \
+               $(shell find $(TRACE_CSR_DIR) -name "*.h")     \
```
poemonsense: Let's see how the headers in plugin is organized. Maybe we need to change how plugin headers are included

Anzooooo:
ok, thanks for the feedback. I'll reorganize the code.

src/test/csrc/verilator/emu.cpp
```diff
@@ -161,6 +167,9 @@ inline EmuArgs parse_args(int argc, const char *argv[]) {
     { "dramsim3-ini",      1, NULL,  0  },
     { "dramsim3-outdir",   1, NULL,  0  },
     { "overwrite-auto",    1, NULL,  0  },
+#ifdef PLUGIN_SIMFRONTEND
```
poemonsense: This should not be wrapped. Otherwise argparser will fail because indices will change

Makefile
```diff
@@ -239,6 +239,14 @@ SIM_CXXFLAGS += -I$(abspath $(PLUGIN_CSRC_DIR)/runahead)
 SIM_CXXFILES += $(shell find $(PLUGIN_CSRC_DIR)/runahead -name "*.cpp")
 endif
 
+# SimFrontend plugin
+ifeq ($(ENABLE_SIMFRONTEND), 1)
```
poemonsense, line 243: How about `ifeq ($(PLUGIN_SIMFRONTEND), 1)`?

poemonsense (commented):
Minor changes

Makefile
Anzooooo: OK.
I still want to keep entering `ENABLE_SIMFRONTEND` in the command line.
Since we need to use `PLUGIN_SIMFRONTEND` in `difftest`, I should write it like this in XiangShan is makefile:

```
# run with sim frontend(ideal frontend)
ifeq ($(ENABLE_SIMFRONTEND),1)
PLUGIN_SIMFRONTEND = 1
override SIM_ARGS += --enable-simfrontend
endif
```

What do you think?

poemonsense: I see. We can keep ENABLE_SIMFRONTEND then

Anzooooo: Okay, thank you for your suggestion.

src/test/csrc/difftest/refproxy.h
Anzooooo: I have some ideas, but considering this modification is not strongly related to the work on `simfrontend`, I will remove this change from this PR and open a separate new PR to discuss this issue.

poemonsense (approved):

PR #683, opened 2025-09-18
fpga-diff: trigger FPGA reset from host xdma

xiaokamikami:


xiaokamikami:
Dependent hardware implementation: https://github.com/OpenXiangShan/env-scripts/pull/50

src/test/vsrc/fpga/xdma_axi4lite_bar.v
```diff
@@ -0,0 +1,109 @@
+module xdma_axi4lite_bar #(
```
poemonsense: Should this file be in env-scripts?

xiaokamikami: ok

src/test/csrc/fpga/fpga_main.cpp
```diff
@@ -79,6 +79,10 @@ void set_diff_ref_so(char *s) {
 
 void fpga_init() {
   xdma_device = new FpgaXdma();
+#ifndef FPGA_SIM
+  xdma_device->fpga_reset_io(true);
+  std::this_thread::sleep_for(std::chrono::nanoseconds(100));
```
klin02: Is there a simpler way to write it, such as usleep?
It is better to keep the code simple and easy to read.

klin02:
This change adds support for writing to FPGA address 0x0 to trigger reset, which enables hardware-side timing (starting when reset ends) to align with software-side timing (difftest_init), producing more accurate performance data.

Note: existing APIs (ddr_workload_load, core_reset, core_restart) do not work for now and are only placeholders for PCIe-based workload loading. These will be fixed in a future update.
https://github.com/OpenXiangShan/difftest/blob/8c528d8e30f8ec072194708a301f41ddfe3bd87c/src/test/csrc/fpga/xdma.h#L90-L99

klin02 (approved):

PR #686, opened 2025-09-20
Uart16550 integration

ARN1954:


poemonsense:
The uart in DiffTest contains only input/output functions. It is not bind to any specific uart model.

We'll not accept such changes as the uart device/driver should be implemented in DUT, not in DiffTest.

poemonsense:
I believe there's no need to change DiffTest if the DUT implements uart16550, uartlite, or any other uart models, because all of them should use only `getc` and `putc` functions through the top-level difftest_uart IOs.

PR #701, opened 2025-10-16
fix(Gateway): explicitly specify the width of `packed`

Tang-Haojin:


Tang-Haojin:
See https://github.com/OpenXiangShan/XiangShan/actions/runs/18550542685/job/52877116736.

Without this, firtool will throw: "error: public module port must have known width".

PR #713, opened 2025-11-07
feat(top): allow extra CPU IO with connectTopIOs

klin02:
This change allows passing some extra CPU IOs as return value of connectTopIOs, and then DiffTest will also expose this IOs to outer SimTop.


poemonsense:
LGTM. Have you tested it with XiangShan?

PR #714, opened 2025-11-08
feat(difftest): support DiffPhyRegState and DiffArchRenameTable

klin02:
This commit introduces DiffPhyRegState and DiffArchRenameTable as new Difftest 
interfaces. A new option `softArchUpdate` is also added (disabled by default).

By default, Difftest replaces PhyReg and RenameTable with ArchReg and CommitData
(WriteBack results) on the hardware side to keep backward compatibility.
When `softArchUpdate` is enabled, this substitution is deferred to the software side,
reducing hardware area overhead by avoiding extra read ports in the register files.

### DIFF
--- a/src/main/scala/Bundles.scala
+++ b/src/main/scala/Bundles.scala
@@ -175,12 +175,12 @@ class TriggerCSRState extends DifftestBaseBundle {
   val tinfo = UInt(64.W)
 }
 
-class ArchRenameTable(numRegs: Int, numPregs: Int) extends DifftestBaseBundle {
-  val value = Vec(numRegs, UInt(log2Ceil(numPregs).W))
+class ArchRenameTable(numRegs: Int, numPhyRegs: Int) extends DifftestBaseBundle {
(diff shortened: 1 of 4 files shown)


poemonsense (approved):

src/test/csrc/difftest/difftest.h
```diff
@@ -397,26 +398,37 @@ class Difftest {
...
+    return dut->commit_data[i].data;
+#elif defined(CONFIG_DIFFTEST_ARCHFPREGSTATE)
+    if (dut->commit[i].fpwen) {
+      return get_fp_data(i);
     } else
-#endif // CONFIG_DIFFTEST_ARCHFPREGSTATE
-      return
-#ifdef CONFIG_DIFFTEST_INTWRITEBACK
-          dut->wb_int[dut->commit[i].wpdest].data;
 #else
```
poemonsense, line 430: This `#else` is wrong. When both Int and Arch are defined, they should exist both.

See https://github.com/OpenXiangShan/difftest/actions/runs/19317148467/job/55250795228

PR #717, opened 2025-11-12
fix(difftest): remove unused vector logic from get_commit_data

klin02:
The get_commit_data function serves two main purposes: (1) synchronizing MMIO operations to NEMU, and
(2) recording per-instruction writeback values into the trace for debugging.

Since vector instructions do not involve MMIO and their writeback widths are incompatible with the existing trace format, the unused vector-related logic has been removed from this function.


poemonsense (approved):

klin02:
> Then do we still need vecwen in InstrCommit?

We still need to use `vec_commit_data` for vecLoad check, just not used in the `get_commit_data` func.

PR #719, opened 2025-11-13
fix(difftest): remove unused idtfr from RefillEvent

klin02:


poemonsense (approved):
@Anzooooo Is this ok? Someone said it's unused now

Anzooooo (approved):
I believe it's safe to remove this unused signal. 
We have indeed never used this signal.

PR #747, opened 2025-12-01
feat(delta): support multi-cycle Delta transmission

klin02:
This update adds multi-cycle transmission support to the Delta, named as Delta Splitting.

Previously, Delta allowed splitting a register file of size N into N bundles, sending only the bundles that changed. This change further splits those N bundles across multiple cycles, reducing the number of elems Batch processing per cycle and thereby lowering hardware area.

Key details:
1. The N bundles are grouped into chunks of size DeltaLimit. A chunk is transmitted only when it contains valid updates. As a result, a single-cycle Delta input may require up to N / DeltaLimit cycles to complete transmission.
2. A new DeltaQueue is introduced. When the previous cycle's data is still in pending, new Delta inputs are buffered in the queue. The queue depth is currently set to 3, which is sufficient under Squash, where data is compressed and multi-ycycle produces only a single Delta input.
3. In DPI-C, a set of Delta data may require multi steps to process. To support this behavior, we move the checking logic (xx_nstep) after each step instead of the end of DPI-C. When the current step is pending, it means incoming data is temporarily stored in delta_buffer; if not pending, we will copy delta_buffer to difftest_state and call xx_nstep for check.

### DIFF
--- a/src/main/scala/DPIC.scala
+++ b/src/main/scala/DPIC.scala
@@ -89,7 +89,7 @@ abstract class DPICBase(config: GatewayConfig) extends ExtModule with HasExtModu
   def getPacketDecl(gen: DifftestBundle, prefix: String, config: GatewayConfig): String = {
     val dut_zone = if (config.hasDutZone) "dut_zone" else "0"
     val dut_index = if (config.isBatch) "dut_index" else "0"
-    val packet = if (config.isDelta && gen.isDeltaElem) {
+    val packet = if (config.isDelta && (gen.isDeltaElem || gen.desiredCppName == "delta_info")) {
       s"DELTA_BUF(${prefix}coreid)->${gen.actualCppName}"
     } else {
       s"DUT_BUF(${prefix}coreid, $dut_zone, $dut_index)->${gen.actualCppName}"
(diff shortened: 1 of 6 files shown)


src/main/scala/DPIC.scala
```diff
@@ -292,34 +292,42 @@ class DPICBatch(template: Seq[DifftestBundle], batchIO: BatchIO, config: Gateway
...
            |    else if (id == BatchStep) {
-           |      ${if (config.isDelta) "dStats->sync(0, dut_index);" else ""}
+           |      $stepPending
            |      dut_index = (dut_index + 1) % CONFIG_DIFFTEST_BATCH_SIZE;
            |#ifdef CONFIG_DIFFTEST_QUERY
            |      difftest_query_step();
            |#endif // CONFIG_DIFFTEST_QUERY
+           |#ifdef CONFIG_DIFFTEST_INTERNAL_STEP
+           |#ifdef FPGA_HOST
+           |      extern void fpga_nstep(uint8_t step);
```
poemonsense, line 324: If this file can be determined purly by C++ macros, move it to C++ please (in a future PR)

poemonsense (approved):
Please check whether C++ files defined in Scala can be moved into `src/test/csrc` and controlled by generated C++ macros.

PR #748, opened 2025-12-01
fix(SimJTAG): remove wire delays for newer version of verilator

cyyself:
In newer versions of Verilator, functions with delays are not legal, which causes a compile error:

```
%Error: /path-to-difftest/src/test/vsrc/common/SimJTAG.v:52:17: Delays are not legal in functions. Suggest use a task (IEEE 1800-2023 13.4.4)
                                                                                           : ... note: In instance 'SimTop'
   52 |    wire         #0.1 __jtag_TDO = jtag_TDO_driven ?
      |                 ^
        ... See the manual at https://verilator.org/verilator_doc.html?v=5.041 for more assistance.
%Error: Exiting due to 1 error(s)
```

As the delay not be useful in verilator, thus remove it.

### DIFF
--- a/src/test/vsrc/common/SimJTAG.v
+++ b/src/test/vsrc/common/SimJTAG.v
@@ -49,7 +49,7 @@ module SimJTAG #(
 
    logic [31:0] random_bits = $random;
 
-   wire         #0.1 __jtag_TDO = jtag_TDO_driven ?
+   wire         __jtag_TDO = jtag_TDO_driven ?
                 jtag_TDO_data : random_bits[0];
 
    bit          __jtag_TCK;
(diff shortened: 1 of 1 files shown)


cyyself:
> I am not sure why we add wire delay for SimJTAG, maybe @poemonsense know about it?

Maybe just copied from rocket-chip: https://github.com/chipsalliance/rocket-chip/blob/master/src/main/resources/vsrc/SimJTAG.v

poemonsense:
The file was brought from https://github.com/chipsalliance/rocket-chip/blob/master/src/main/resources/vsrc/SimJTAG.v.

Which verilator version are you using? I'm using v5.042 and it works.

If this is a serious issue, we may need to fix in the upstream first to ensure its correctness, given that we don't know why

PR #750, opened 2025-12-01
feat(ci): add test for build EMU with all bundles

poemonsense:
The DifftestInterfaces Module is expected to include all DiffTest bundles for testing. This commit adds a EMU test that builds the executable with all interfaces. It should check whether all bundles and their corresponding checkers can be successfully compiled.


src/test/scala/DifftestTop.scala
```diff
@@ -23,42 +23,51 @@ import scala.annotation.tailrec
...
+  val hcsr_state = DifftestModule(new DiffHCSRState, dontCare = true)
+  val debug_mode = DifftestModule(new DiffDebugMode, dontCare = true)
+  val trigger_csr_state = DifftestModule(new DiffTriggerCSRState, dontCare = true)
+  val int_writeback = DifftestModule(new DiffIntWriteback, dontCare = true)
+  val fp_writeback = DifftestModule(new DiffFpWriteback, dontCare = true)
+  val vec_writeback = DifftestModule(new DiffVecWriteback, dontCare = true)
+  val vec_v0_writeback = DifftestModule(new DiffVecV0Writeback, dontCare = true)
+  val arch_int_reg_state = DifftestModule(new DiffArchIntRegState, dontCare = true)
+  val arch_fp_reg_state = DifftestModule(new DiffArchFpRegState, dontCare = true)
+  val arch_vec_state = DifftestModule(new DiffArchVecRegState, dontCare = true)
```
klin02, line 41: vec_load is fixed. And we can use `PhyReg+RenameTable` instead of `ArchReg+Writeback`.

klin02 (approved):

PR #751, opened 2025-12-02
fix(difftest): refine vec_load checkers with common param

klin02:
In #744, we seperate load checkers for Squash, and unify the checking logic into a function with common param list.

This change further unifies common params for vec_load checker, and call it accordingly in both Squash and non-Squash load checker.


klin02:
Test well with #750

PR #754, opened 2025-12-04
refactor(difftest): remove DiffWriteback and privatize DiffArchReg

klin02:
This change removes the legacy writeback API and marks DiffArchReg along with other internal Difftest interfaces as private to prevent external use.

When DiffPhyReg has the same numRegs as DiffArchReg, it is now internally converted to DiffArchReg to maintain consistent behavior.

Users should replace DiffArchReg + DiffWriteBack with one of the following:
  (a) DiffPhyReg (when numRegs match), or
  (b) DiffPhyReg + DiffRenameTable.


src/test/scala/DifftestTop.scala
```diff
@@ -32,22 +32,14 @@ class DifftestInterfaces extends Module {
...
-  // val arch_vec_rename_table = DifftestModule(new DiffArchVecRenameTable(64), dontCare = true)
-  // val phy_int_reg_state = DifftestModule(new DiffPhyIntRegState(192), dontCare = true)
-  // val phy_fp_reg_state = DifftestModule(new DiffPhyFpRegState(128), dontCare = true)
-  // val phy_vec_reg_state = DifftestModule(new DiffPhyVecRegState(64), dontCare = true)
+  val arch_int_rename_table = DifftestModule(new DiffArchIntRenameTable(224), dontCare = true)
+  val arch_fp_rename_table = DifftestModule(new DiffArchFpRenameTable(192), dontCare = true)
+  val arch_vec_rename_table = DifftestModule(new DiffArchVecRenameTable(150), dontCare = true)
+  val phy_int_reg_state = DifftestModule(new DiffPhyIntRegState(224), dontCare = true)
+  val phy_fp_reg_state = DifftestModule(new DiffPhyFpRegState(192), dontCare = true)
+  val phy_vec_reg_state = DifftestModule(new DiffPhyVecRegState(150), dontCare = true)
```
klin02, line 42: Previous `numPregs` of vec is same as `ArchVecReg`, requiring VecRenameTable is empty.
We now modify `numPregs` as https://github.com/OpenXiangShan/XiangShan/blob/master/src/main/scala/xiangshan/Parameters.scala.

README.md
```diff
@@ -140,25 +140,18 @@ coherence via RefillTest.
 | `DiffArchEvent` | Exceptions and interrupts | Yes |
 | `DiffInstrCommit` | Executed instructions | Yes |
 | `DiffTrapEvent` | Simulation environment call | Yes |
-| `DiffArchIntRegState` | General-purpose registers | Yes |
-| `DiffArchFpRegState` | Floating-point registers | No |
-| `DiffArchVecRegState` | Vector registers | No |
 | `DiffCSRState` | Control and status registers (CSRs) | Yes |
+| `DiffPhyIntRegState`<br>~~`(DiffArchIntRegState)`~~ | Physical general-purpose registers<br>(=>`DiffArchIntRegState` when `numPhyRegs = 32`) | Yes |
```
poemonsense: We can explicitly explain the design here instead of using `=>`. This helps others understand it

Or, we can add a paragraph describing the detailed design after the table.

poemonsense (approved):

README.md
klin02: Maybe replace `=>` with `Act as`?

poemonsense: This would be better

PR #756, opened 2025-12-04
feat(verilator,gsim): add PGO_BOLT option

cyyself:
The PGO data generated by hardware branch tracing directly with nearly zero runtime overhead to the already compiled binary, thus making the PGO build really quick (~1min), also avoiding the need for a full recompilation. This results in a much faster build process while still benefiting from the performance improvements provided by PGO.

The results show that it only takes 1:27 with  `PGO_BOLT=1` and `PGO_MAX_CYCLE=100000` to build the Verilator emulator with PGO, compared to 10+ minutes traditionally, and the performance is nearly the same, both finished CoreMark in ~26s on my 13900K with LLVM 22, while the non-PGO build takes 84s.

This process requires Linux-perf to collect the profile data, and BOLT to apply the optimizations. When profiling with Linux-perf, please ensure that the system has set `sysctl -w kernel.perf_event_paranoid=-1` to allow perf to collect the necessary data.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   test-difftest-main:
     runs-on: ubuntu-latest
-    container: ghcr.io/openxiangshan/xs-env:ubuntu-22.04
+    container: ghcr.io/openxiangshan/xs-env:ubuntu-24.04
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -71,10 +71,18 @@ jobs:
             ./build/emu --no-diff -C 10000
             make clean
 
-      - name: Basic Difftest with PGO build
+      - name: Basic Difftest with PGO build using compiler instrumentation
+        run: |
+            cd $NOOP_HOME
+            sudo ln -s $(realpath $(dirname $(readlink -f $(which llvm-bolt)))/../lib/libbolt_rt_instr.a) /usr/lib/ || true # workaround for Debian / Ubuntu packaging issue
```
cyyself: Here is a bug in Debian's packaging. The downstream Ubuntu was also affected. I have reported here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1121914

PR #769, opened 2025-12-13
fix(night-ci): move fpga_diff synth only for evening

klin02:


.github/workflows/fpga.yml
```diff
@@ -66,8 +62,7 @@ jobs:
           echo "VIVADO_HIER_MD=$MD_PATH" >> "$GITHUB_ENV"
 
 
-      - name: Comment utilization to PR
-        if: ${{ github.event_name == 'pull_request' }}
+      - name: Comment utilization to latest PR on master        if: ${{ github.ref == 'refs/heads/master' }}
```
poemonsense: Is this `if` required? nightly tests only run on latest master

.github/workflows/fpga.yml
```diff
@@ -81,4 +76,33 @@ jobs:
...
+              return;
+            }
+
+            await github.rest.issues.createComment({
+              owner,
+              repo,
+              issue_number: pr.number,
+              body,
+            });
+            core.info(`Commented utilization to PR #${pr.number}`);
```
poemonsense, line 122: Will this comment be duplicated? Nightly test runs every night but we didn't check whether the comment has been written to the corresponding PR. If it's already written, we should not write it the second time

PR #775, opened 2025-12-17
feat(difftest): separate checkers into multiple files

poemonsense:
We move different checkers to their own files to improve the code readability. By introducing the DiffTestChecker class, various checkers can be used in the same way and executed one by one.

The checkers are put under the `checkers` directory with each cpp file containing a set of checkers for a single verification target. Currently the targets include instruction commits, traps, TLB, load instructions, store instructions, global memory update, cache refill, and some misc state synchronizations.

### DIFF
--- a/src/test/csrc/difftest/difftest.cpp
+++ b/src/test/csrc/difftest/difftest.cpp
@@ -504,7 +504,8 @@ inline int Difftest::check_all() {
 #ifdef FUZZING
     if (in_disambiguation_state()) {
       Info("Mismatch detected with a disambiguation state at pc = 0x%lx.\n", dut->trap.pc);
-      state->raise_trap(STATE_FUZZ_COND) return DiffTestChecker::STATE_TRAP;
+      state->raise_trap(STATE_FUZZ_COND);
+      return DiffTestChecker::STATE_TRAP;
     }
 #endif
 #ifdef FUZZER_LIB


poemonsense:
I've separated the lengthy `difftest.cpp` file into multiple checkers, each of which contains the checker logic for each bundle  or a specific area. @Anzooooo Does this look ok to you?

klin02 (dismissed):
LGTM. Should I open a seperate PR in XiangShan to test this change first?

Anzooooo:
I believe it's highly likely to be fine, but I might need a little more time to review it. ) :kissing_cat:

src/test/csrc/difftest/difftest.cpp
```diff
@@ -491,28 +497,23 @@ inline int Difftest::check_all() {
...
-    return 1;
+    return DiffTestChecker::STATE_DIFF;
   }
 
   if (proxy->compare(dut) || pc_mismatch) {
 #ifdef FUZZING
     if (in_disambiguation_state()) {
       Info("Mismatch detected with a disambiguation state at pc = 0x%lx.\n", dut->trap.pc);
-      return 0;
+      state->raise_trap(STATE_FUZZ_COND) return DiffTestChecker::STATE_TRAP;
```
klin02: Maybe the `return` should be in a new line?

poemonsense: Fixed

PR #784, opened 2025-12-25
fix(Delta): validate DeltaInfo only when deltas valid

klin02:


klin02:
After this change, we can test XiangShan with DeltaQueue depth 4 first before Pipeline refactor ready.

poemonsense (approved):

PR #787, opened 2025-12-29
feat(fpga_diff): support async_clock for Diff2AXI

xiaokamikami:
Use asynchronous FIFO to adapt the transmission function of difftest packages with different frequencies

On the nutshell board test, the simulation speed is: 38015.93 KHz

### DIFF
--- a/src/main/scala/fpga/Host.scala
+++ b/src/main/scala/fpga/Host.scala
@@ -45,14 +45,8 @@ class Difftest2AXIs(val difftest_width: Int, val axis_width: Int) extends Module
 
   // Write clock domain (using default clock)
   val wr_ptr = RegInit(0.U(fifo_addr_width.W))
-  val rd_cnt_sync = RegInit(0.U(64.W))
-  val wr_occupancy = Reg(UInt(64.W)) // Occupancy in write clock domain
-
-  // Synchronize read counter to write clock domain
-  rd_cnt_sync := Delayer(rd_cnt, 2)
-
-  // Calculate occupancy (in write clock domain)
-  wr_occupancy := wr_cnt - rd_cnt_sync
+  val rd_cnt_sync = Delayer(rd_cnt, 2) // Synchronize read counter to write clock domain
+  val wr_occupancy = wr_cnt - rd_cnt_sync // Calculate occupancy in write clock domain
 
   // Write side (difftest side)
   val pktID = RegInit(0.U(pkt_id_w.W))
@@ -77,14 +71,8 @@ class Difftest2AXIs(val difftest_width: Int, val axis_width: Int) extends Module
   // Read clock domain
   withClock(io.pcie_clock) {
     val rd_ptr = RegInit(0.U(fifo_addr_width.W))
-    val wr_cnt_sync = RegInit(0.U(64.W))
-    val rd_occupancy = Reg(UInt(64.W)) // Occupancy in read clock domain
-
-    // Synchronize write counter to read clock domain
-    wr_cnt_sync := Delayer(wr_cnt, 2)
-
-    // Calculate occupancy (in read clock domain)
-    rd_occupancy := wr_cnt_sync - rd_cnt
+    val wr_cnt_sync = Delayer(wr_cnt, 2) // Synchronize write counter to read clock domain
+    val rd_occupancy = wr_cnt_sync - rd_cnt // Calculate occupancy in read clock domain
 
     val inTransfer = RegInit(false.B)
     val mix_data = RegInit(0.U((difftest_width + pkt_id_w).W))
@@ -107,7 +95,7 @@ class Difftest2AXIs(val difftest_width: Int, val axis_width: Int) extends Module
         inTransfer := true.B
         sendCnt := 0.U
       }
-    }.elsewhen(inTransfer) {
+    }.otherwise {
       when(io.axis.fire) {
         when(sendLast) {
           sendCnt := 0.U


src/test/csrc/fpga_sim/xdma_sim.cpp
```diff
@@ -106,8 +106,10 @@ class xdma_sim {
     if (shm_ptr->write_size >= shm_ptr->read_size) {
       assert(tlast == 1); // Check if tlast set properly
       shm_ptr->read_waiting = false;
+      //printf("[XDMA] Write size: %lu ;Read size: %lu\n", shm_ptr->write_size,  shm_ptr->read_size);
```
klin02: Redundant printf

src/main/scala/fpga/Host.scala
```diff
@@ -20,119 +20,130 @@ import chisel3._
...
+  // Write clock domain (using default clock)
+  val wr_ptr = RegInit(0.U(fifo_addr_width.W))
+  val rd_cnt_sync1 = RegInit(0.U(64.W))
+  val rd_cnt_sync2 = RegInit(0.U(64.W))
+  val wr_occupancy = Reg(UInt(64.W)) // Occupancy in write clock domain
 
-  val inTransfer = RegInit(false.B)
+  // Synchronize read counter to write clock domain
+  rd_cnt_sync1 := RegNext(rd_cnt, 0.U)
+  rd_cnt_sync2 := RegNext(rd_cnt_sync1, 0.U)
```
klin02: We have a `Delayer` API. You can simply use the API like `val rd_cyt_sync = Delayer(rd_cnt, 2)

src/main/scala/fpga/Host.scala
```diff
@@ -20,119 +20,130 @@ import chisel3._
...
-    }
-    ram.read(rd.addr)
-  }
-  rd.data := Cat(rd_data_vec.reverse)
-}
-
 class Difftest2AXIs(val difftest_width: Int, val axis_width: Int) extends Module {
   val io = IO(new Bundle {
-    val clock = Input(Clock())
+    val rd_clock = Input(Clock()) // Read clock
```
klin02: Maybe directly use pcie_clock? It seems to be more clear.

src/main/scala/fpga/Host.scala
```diff
@@ -20,119 +20,130 @@ import chisel3._
...
+    fifo_ram.write(wr_ptr, io.difftest.data)
+    wrRangeCounter := Mux(wrRangeCounter === (numPacketPerRange - 1).U, 0.U, wrRangeCounter + 1.U)
+    wr_cnt := wr_cnt + 1.U // Increment write counter
   }
 
-  // Backpressure
-  io.clock_enable := range_vnum =/= numRange.U
+  // Increment write pointer
+  when(wr_en) {
+    wr_ptr := wr_ptr + 1.U
```
klin02: We can also move this line to line 65~68, where we handle ram write logic.

src/main/scala/fpga/Host.scala
```diff
@@ -20,119 +20,130 @@ import chisel3._
...
+  }
 
-  // Write
-  when(buf_wen) {
-    wrPkt := Mux(wrPkt === (totalPacket - 1).U, 0.U, wrPkt + 1.U)
+  // Increment packet ID when starting a new range
+  when(wr_en && wrRangeCounter === (numPacketPerRange - 1).U) {
```
klin02: Also move the whenContext inside `when (wr_en)` in line65~68, like:
```
  when(wr_en) {
    fifo_ram.write(wr_ptr, io.difftest.data)
    wr_ptr := wr_ptr + 1.U
    wrRangeCounter := Mux(wrRangeCounter === (numPacketPerRange - 1).U, 0.U, wrRangeCounter + 1.U)
    when(wrRangeCounter === (numPacketPerRange - 1).U) {
       pktID := pktID + 1.U
    }
    wr_cnt := wr_cnt + 1.U // Increment write counter
  }
```
That is to move all updating logic inside `wr_en` condition.

src/main/scala/fpga/Host.scala
```diff
@@ -20,119 +20,130 @@ import chisel3._
...
+        rd_ptr := rd_ptr + 1.U
+        rd_cnt := rd_cnt + 1.U // Increment read counter
+        // counter := 0.U  // 0~7
+        inTransfer := true.B
         sendCnt := 0.U
-        rdPkt := nextRdPkt
-        when(rdRangeLast) {
-          inTransfer := false.B
+      }
+    }.elsewhen(inTransfer) {
```
klin02: when(!inTranser){...}.otherwise{...}

src/main/scala/fpga/Host.scala
```diff
@@ -144,11 +155,27 @@ class HostEndpoint(
...
+
+  // Handle cross-clock domain data transfer
+  // Difftest input domain (local clock)
+  withClock(clock) {
+    diff2axis.io.difftest := io.difftest
+  }
+
+  // AXI-Stream output domain (PCIe clock)
+  withClock(io.pcie_clock) {
+    io.to_host_axis <> diff2axis.io.axis
```
klin02: Should we add `withClock` for wiring?

src/main/scala/fpga/Host.scala
```diff
@@ -19,120 +19,122 @@ package difftest.fpga
...
-    ram.read(rd.addr)
-  }
-  rd.data := Cat(rd_data_vec.reverse)
-}
+import difftest.util.Delayer
 
 class Difftest2AXIs(val difftest_width: Int, val axis_width: Int) extends Module {
   val io = IO(new Bundle {
-    val clock = Input(Clock())
+    val pcie_clcok = Input(Clock()) // Read clock
```
klin02: clock instead of clcok

src/main/scala/fpga/Host.scala
```diff
@@ -144,11 +146,20 @@ class HostEndpoint(
...
     val clock_enable = Output(Bool())
+    val pcie_clock = Input(Clock())
   })
+
+  // Instantiate the converter module
   val diff2axis = Module(new Difftest2AXIs(diffWidth, axisWidth))
-  diff2axis.io.clock := clock
+
+  // Connect clock and reset signals
+  diff2axis.io.pcie_clcok := io.pcie_clock
```
klin02: clock instead of clcok

src/main/scala/fpga/Host.scala
```diff
@@ -19,120 +19,122 @@ package difftest.fpga
...
+  io.clock_enable := fifo_not_full
 
-  // Write
-  when(buf_wen) {
-    wrPkt := Mux(wrPkt === (totalPacket - 1).U, 0.U, wrPkt + 1.U)
-  }
+  // Read clock domain
+  withClock(io.pcie_clcok) {
+    val rd_ptr = RegInit(0.U(fifo_addr_width.W))
+    val wr_cnt_sync = RegInit(0.U(64.W))
```
klin02: `val wr_cnt_sync = Delayer(wr_cnt, 2)`
Assign something to a Reg will add one extra delay, same to rd_cnt_sync

src/main/scala/fpga/Host.scala
```diff
@@ -19,120 +19,122 @@ package difftest.fpga
...
+    val rd_occupancy = Reg(UInt(64.W)) // Occupancy in read clock domain
```
klin02: Simplify to `val rd_occupacy = wr_cnt_sync - rd_cnt`.

src/main/scala/fpga/Host.scala
klin02: Use `otherwise` instead of `else when(inTransfer)`

klin02 (approved):

PR #790, opened 2025-12-30
feat(difftest): support module prefix for SimTop

klin02:


src/main/scala/SimTop.scala
```diff
@@ -90,7 +90,8 @@ trait HasDiffTestInterfaces extends ImplicitClock with ImplicitReset { this: Raw
 }
 
 // Top-level module for DiffTest simulation. Will be created by DifftestModule.top
-class SimTop[T <: RawModule with HasDiffTestInterfaces](cpuGen: => T) extends Module {
+class SimTop[T <: RawModule with HasDiffTestInterfaces](cpuGen: => T, modPrefix: Option[String]) extends Module {
+  override def desiredName: String = modPrefix.getOrElse("") + super.desiredName
```
poemonsense, line 94: should be implemented as an annotation in the future

See https://github.com/poemonsense/SVM/blob/main/src/test/scala/NestedPrefixModulesAnnotation.scala

poemonsense (approved):

PR #793, opened 2025-12-31
fix(difftest): invalidate commit when exception/interrupt

klin02:


poemonsense (approved):
Approved for now.

Should check why and remove it in the future

PR #796, opened 2026-01-06
fix(difftest): fix xx-check memory overflow

xiaokamikami:
When handling the allocated memory for the application when exiting the program


src/test/csrc/difftest/difftest.cpp
```diff
@@ -198,6 +198,27 @@ Difftest::Difftest(int coreid) {
 }
 
 Difftest::~Difftest() {
+  for (auto checker : checkers) {
+    delete checker;
+  }
+  checkers.clear();
```
poemonsense: I think `checkers.clear();` is not required.

poemonsense (approved):

PR #797, opened 2026-01-06
feat(difftest): decouple clockgate from fpga_sim

klin02:
To prepare for upcoming pipeline refactoring, this change decouples clockgate
and ref_clock from fpga_sim. In addition to FPGA AXI, clockgate can now also be
reused by the Delta, enabling Difftest to process data from a single DUT cycle
across multiple cycles.


src/main/scala/Difftest.scala
```diff
@@ -789,4 +792,28 @@ object DifftestModule {
     difftestV += s"`define CPU_$cpu_s"
     FileControl.write(difftestV, "DifftestMacros.svh")
   }
+
+  def generateClockGate(): Unit = {
```
poemonsense: For verilog modules whose design do not depend on run-time arguments, they should be put in `src` directory.

poemonsense (commented):
Please let @xiaokamikami have a review. I may not have enough time to review the details in the future

Suggestion:

Since now we are defining the SimTop in Chisel, some logic in `top.v` should be moved into Chisel under SimTop, especially the module connections. For example, the clock and clock gate can be moved into Chisel.

src/main/scala/Difftest.scala
klin02: We originally put clockgate under `generate` to simplify include-path management across different build configs. It is used not only by FPGA/FPGA_SIM but also when Delta is enabled, so it may be hard to give it an appropriate path under `src`.

In addition, files under `src` need to be explicitly copied into fpga-release, it may be simplified when put it in `generate`.

xiaokamikami (dismissed):

xiaokamikami (approved):

PR #799, opened 2026-01-09
fix(difftest): support do_check for LoadSquashChecker

klin02:
This change moves do_load_check and do_vec_load_check into different checker
classes depending on whether Squash is enabled.

We use macros instead of shared base class for the following reasons:
1) the two checker classes never use the check functions at the same time, and
2) introducing a common base would require repeated class derivation in
   Difftest, leading to multiple virtual derivations and additional
   initialization overhead.


poemonsense (commented):
Let Chenxi Yan to review it please

CrystalCrack (approved):
LGTM

PR #809, opened 2026-01-20
fix(ci): resolve compilation errors from framework changes

CrystalCrack:
- Rename variables and functions to match new framework APIs
- Add missing includes.


.github/workflows/nightly.yml
```diff
@@ -69,7 +69,7 @@ jobs:
             cp -r $GITHUB_WORKSPACE .
             make bootrom CROSS=riscv64-linux-gnu-
             make emu XFUZZ=1 -j2                                           \
-              REF=$SPIKE_HOME/difftest/build/riscv64-spike-so LLVM_COVER=1 \
+              REF=Spike LLVM_COVER=1 \
```
poemonsense: Is this required?

CrystalCrack: The chosen of REF model is defined here:

https://github.com/OpenXiangShan/difftest/blob/7cd70cc49c468577d97b8bed7de125951a5e1ce7/Makefile#L187

the default paths for NEMU and Spike are declared in `refproxy.cpp`:

https://github.com/OpenXiangShan/difftest/blob/7cd70cc49c468577d97b8bed7de125951a5e1ce7/src/test/csrc/difftest/refproxy.cpp#L318-L325

So I think it doesn't work to put the path into `REF`. It must be set to "Spike" to function correctly.

poemonsense: I think the original implementation is good because we need LLVM_COVER=1 (for coverage instrumentation).

In the Makefile, there is also one line `SIM_CXXFLAGS += -DREF_PROXY=LinkedProxy -DLINKED_REFPROXY_LIB=\\\"$(REF)\\\"` which will be used here

PR #813, opened 2026-01-23
feat(difftest): checkers profiling

CrystalCrack:
This PR introduces a performance measurement feature for Difftest checkers, allowing developers to track and report the time spent in each checker when the new `DIFFTEST_CHECKER_PERF` option is enabled.

### DIFF
--- a/src/test/csrc/common/stopwatch.cpp
+++ b/src/test/csrc/common/stopwatch.cpp
@@ -1,10 +1,26 @@
+/***************************************************************************************
+* Copyright (c) 2020-2026 Institute of Computing Technology, Chinese Academy of Sciences
+* Copyright (c) 2026 Beijing Institute of Open Source Chip
+*
+* DiffTest is licensed under Mulan PSL v2.
+* You can use this software according to the terms and conditions of the Mulan PSL v2.
+* You may obtain a copy of Mulan PSL v2 at:
+*          http://license.coscl.org.cn/MulanPSL2
+*
+* THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND,
+* EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+* MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+*
+* See the Mulan PSL v2 for more details.
+***************************************************************************************/
+
 #include "stopwatch.h"
 #include <algorithm>
 #include <iomanip>
 #include <iostream>
 #include <map>
 
-std::vector<Stopwatch *> all_stopwatches;
+static std::vector<Stopwatch *> all_stopwatches;
 std::vector<std::string> StopwatchTypeNames = {"Difftest Checkers", "Others"};
 
 Stopwatch::Stopwatch(std::string name, StopwatchType type) : name(std::move(name)), type(type) {
@@ -65,7 +81,7 @@ void Stopwatch::print_stats(StopwatchType type) {
 
   size_t type_index = static_cast<size_t>(type);
   std::string type_name = (type_index < StopwatchTypeNames.size()) ? StopwatchTypeNames[type_index] : "Unknown";
-  std::cout << "\n===== DiffTest " << type_name << " Performance Stats =====" << std::endl;
+  std::cout << "\n===== " << type_name << " Performance Stats =====" << std::endl;
   std::cout << std::left << std::setw(40) << "Name" << "Time (ms)" << std::endl;
   std::cout << std::string(60, '-') << std::endl;
 
--- a/src/test/csrc/common/stopwatch.h
+++ b/src/test/csrc/common/stopwatch.h
@@ -1,3 +1,19 @@
+/***************************************************************************************
+* Copyright (c) 2020-2026 Institute of Computing Technology, Chinese Academy of Sciences
+* Copyright (c) 2026 Beijing Institute of Open Source Chip
+*
+* DiffTest is licensed under Mulan PSL v2.
+* You can use this software according to the terms and conditions of the Mulan PSL v2.
+* You may obtain a copy of Mulan PSL v2 at:
+*          http://license.coscl.org.cn/MulanPSL2
+*
+* THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND,
+* EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+* MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+*
+* See the Mulan PSL v2 for more details.
+***************************************************************************************/
+
 #ifndef __STOPWATCH_H
 #define __STOPWATCH_H
 
--- a/src/test/csrc/difftest/checkers.h
+++ b/src/test/csrc/difftest/checkers.h
@@ -20,23 +20,20 @@
 #include "common.h"
 #include "diffstate.h"
 #include "refproxy.h"
+#include "stopwatch.h"
 
 #ifdef CONFIG_DIFFTEST_CHECKER_PERF
-#include "stopwatch.h"
 #include <cxxabi.h>
 #endif
 
 class DiffTestChecker {
 public:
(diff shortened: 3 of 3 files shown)


src/test/csrc/difftest/difftest.cpp
```diff
@@ -31,6 +31,9 @@
 #include "query.h"
 #endif // CONFIG_DIFFTEST_QUERY
 
+#ifdef CONFIG_DIFFTEST_CHECKER_PERF
+std::vector<DiffTestChecker*> DiffTestChecker::all_checkers;
```
poemonsense: This is not needed because we have a static member in DiffTestCheckers. Right?

src/test/csrc/difftest/checkers.h
```diff
@@ -21,12 +21,82 @@
...
-  DiffTestChecker(DiffState *state, RefProxy *proxy) : state(state), proxy(proxy) {}
+  DiffTestChecker(DiffState *state, RefProxy *proxy) : state(state), proxy(proxy){
+#ifdef CONFIG_DIFFTEST_CHECKER_PERF
+    all_checkers.push_back(this);
+#endif
+  }
   virtual ~DiffTestChecker() = default;
 
-  virtual int step() = 0;
+  int step() {
```
poemonsense: I think we can define two functions (before_step, after_step, or other names) to wrap these code.

Then the `step` function can be implemented as `before_step`, `do_step` and `after_step`, which looks clearer.

src/test/csrc/difftest/checkers.h
```diff
@@ -21,12 +21,82 @@
...
+#ifdef CONFIG_DIFFTEST_CHECKER_PERF
+    if (!name_init) {
+      int status;
+      char* realname = abi::__cxa_demangle(typeid(*this).name(), 0, 0, &status);
```
poemonsense: Can this `name_init` be done in the constructor?

src/test/csrc/difftest/checkers.h
```diff
@@ -21,12 +21,82 @@
...
+      std::sort(sorted_stats.begin(), sorted_stats.end(), [](const Pair& a, const Pair& b){
+          return a.second > b.second;
+      });
+
+      std::cout << "\n===== DiffTest Checker Performance Stats =====" << std::endl;
+      std::cout << std::left << std::setw(40) << "Checker Name" << "Time (us)" << std::endl;
+      std::cout << std::string(60, '-') << std::endl;
+
+      for (const auto& p : sorted_stats) {
+          std::cout << std::left << std::setw(40) << p.first
```
poemonsense: Use two whitespaces for indent (following the style of other code).

src/test/csrc/difftest/checkers.h
```diff
@@ -21,12 +21,82 @@
...
   virtual ~DiffTestChecker() = default;
 
-  virtual int step() = 0;
+  int step() {
+#ifdef CONFIG_DIFFTEST_CHECKER_PERF
+    if (!name_init) {
+      int status;
+      char* realname = abi::__cxa_demangle(typeid(*this).name(), 0, 0, &status);
+      name = (status == 0) ? std::string(realname) : std::string(typeid(*this).name());
+      free(realname);
```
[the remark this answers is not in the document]
poemonsense: Run `make format` before commit @CrystalCrack

src/test/csrc/difftest/difftest.cpp
CrystalCrack: I believe this is required. The line in the `checkers.h` is only a declaration. We still need this definition in the .cpp file to allocate storage for the static member. Otherwise, the linker would report "undefined symbol" error.

image

src/test/csrc/difftest/checkers.h
CrystalCrack: Inside the base class constructor, the object's dynamic type is DiffTestChecker, not the derived class. Therefore, we must defer the name initialization until the object is fully constructed.

src/test/csrc/difftest/checkers.h
```diff
@@ -21,12 +21,51 @@
...
+#include <cxxabi.h>
+#endif
+
 class DiffTestChecker {
 public:
-  DiffTestChecker(DiffState *state, RefProxy *proxy) : state(state), proxy(proxy) {}
-  virtual ~DiffTestChecker() = default;
+  DiffTestChecker(DiffState *state, RefProxy *proxy) : state(state), proxy(proxy) {
+#ifdef CONFIG_DIFFTEST_CHECKER_PERF
+    timer = nullptr;
```
poemonsense: This can be replaced by `  Stopwatch *timer = nullptr;` in line 79

src/test/csrc/difftest/checkers.h
```diff
@@ -21,12 +21,51 @@
...
+#endif
+  }
+  virtual ~DiffTestChecker() {
+#ifdef CONFIG_DIFFTEST_CHECKER_PERF
+    delete timer;
```
poemonsense: only delete it when `if (timer)` and we do not need to check `CONFIG_DIFFTEST_CHECKER_PERF`

src/test/csrc/difftest/checkers.h
```diff
@@ -36,14 +75,17 @@ class DiffTestChecker {
 protected:
   DiffState *state;
   RefProxy *proxy;
+#ifdef CONFIG_DIFFTEST_CHECKER_PERF
```
poemonsense: There's no need to wrap this with `CONFIG_DIFFTEST_CHECKER_PERF`. We can initialize it to `nullptr`

src/test/csrc/common/stopwatch.cpp
```diff
@@ -0,0 +1,78 @@
...
+  return std::chrono::duration<double, std::milli>(total_duration).count();
+}
+
+void Stopwatch::print_stats(StopwatchType type) {
+  std::map<std::string, double> stats;
+  double total_sum = 0;
+
+  for (auto *c: all_stopwatches) {
+    auto time = c->elapsed_ms();
+    if (c->type == type && time > 0) {
```
poemonsense, line 70: Should we check `time > 0`?

src/test/csrc/common/stopwatch.cpp
```diff
@@ -0,0 +1,78 @@
...
+  std::cout << "\n===== DiffTest " << type_name << " Performance Stats =====" << std::endl;
+  std::cout << std::left << std::setw(40) << "Name" << "Time (ms)" << std::endl;
+  std::cout << std::string(60, '-') << std::endl;
+
+  for (const auto &p: sorted_stats) {
+    std::cout << std::left << std::setw(40) << p.first << p.second << std::endl;
+  }
+  std::cout << std::string(60, '-') << std::endl;
+  std::cout << std::left << std::setw(40) << "TOTAL" << total_sum << std::endl;
+  std::cout << "==============================================\n" << std::endl;
```
poemonsense, line 93: Please add one PR comment to show the example output of the stats

src/test/csrc/common/stopwatch.h
```diff
@@ -0,0 +1,40 @@
+#ifndef __STOPWATCH_H
```
poemonsense, line 17: Add license header here please (ICT + BOSC, 2026)

src/test/csrc/common/stopwatch.cpp
```diff
@@ -0,0 +1,78 @@
+#include "stopwatch.h"
```
poemonsense, line 17: Add license header here please (ICT + BOSC, 2026)

src/test/csrc/common/stopwatch.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <algorithm>
+#include <iomanip>
+#include <iostream>
+#include <map>
+
+std::vector<Stopwatch *> all_stopwatches;
```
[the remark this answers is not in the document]
poemonsense: I think we also need to mark this variable as `static`.

src/test/csrc/common/stopwatch.cpp
CrystalCrack: It looks like this
image

src/test/csrc/common/stopwatch.cpp
CrystalCrack: This can ensures that only stopwatches with actual runtime are recorded, avoiding noise in the final statistics.

PR #852, opened 2026-04-10
fix(fpga): handle XDMA sim shutdown and single-channel fallback in software path

xiaokamikami:
## Background

In `FPGA_SIM + USE_THREAD_MEMPOOL + CONFIG_DMA_CHANNELS==1`, the XDMA software path may not exit cleanly or drain the last difftest packets reliably, which can leave the run without a visible `HIT GOOD TRAP`.

## Changes

- add interrupt/close handling for `xdma_sim`
- make XDMA receive threads exit cleanly on shutdown
- add a single-channel fallback path in FPGA sim threaded mode
- switch software reset control to the current host I/O interface

### DIFF
--- a/src/test/csrc/fpga/xdma.cpp
+++ b/src/test/csrc/fpga/xdma.cpp
@@ -142,6 +142,11 @@ void FpgaXdma::start_transmit_thread() {
 }
 
 void FpgaXdma::stop_thansmit_thread() {
+#ifdef FPGA_SIM
+  for (int i = 0; i < CONFIG_DMA_CHANNELS; i++) {
+    xdma_sim_interrupt(i);
+  }
+#endif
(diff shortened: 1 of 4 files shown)


src/test/csrc/fpga/xdma.cpp
```diff
@@ -163,18 +168,50 @@ void FpgaXdma::read_xdma_thread(int channel) {
...
+      break;
+    }
     if (xdma_mempool.write_free_chunk(mem[0], mem_get_idx) == false) {
       printf("It should not be the case that no available block can be found\n");
       assert(0);
     }
   }
 }
 
+void FpgaXdma::read_and_process_fallback() {
```
klin02, line 189: Merge this function with similar `read_and_process()`

PR #861, opened 2026-04-21
feat(sim): add native split view for uart and host logs

xiaokamikami:
## Background

UART output and host-side logs were previously mixed in the same terminal stream, making it difficult to distinguish device UART output from DiffTest and host messages during FPGA/simulation debugging.

## Changes

This PR adds native split view support for separating UART output from host logs:

- Adds a native terminal split view with separate UART and host log panes
- Routes UART output through a dedicated split view UART sink
- Keeps host logs on the regular stdout/stderr path while capturing them into the host pane
- Enables split view support by default for FPGA host builds
- Preserves the mainline event-driven serial-port stop-pipe logic
- Adds SIGINT and shutdown cleanup for restoring terminal state
- Mirrors both streams to durable log files asynchronously:
  - default: `build/logs/uart.log`
  - default: `build/logs/host.log`
- Supports custom log output directories via:
  - `--log=PATH`
  - `log=PATH`
  - `+log=PATH`

When a custom path is provided, the runtime writes:

- `PATH/uart.log`
- `PATH/host.log`

## Verification

- Rebased onto latest `origin/master`
- Squashed into a single PR commit
- Resolved conflicts with recent FPGA serial-port changes
- `git diff --check origin/master..HEAD` passes
- C++ syntax checked with `g++ -std=c++20 -fsyntax-only` for:
  - `splitview.cpp`
  - `args.cpp`
  - `common.cpp`
  - `uart_sink.cpp`
  - `serial_port.cpp`

## Notes

Full `fpga-host` / `emu` build was not run in this checkout because the generated header `difftest-state.h` is absent from the local generated-src environment.


src/test/vsrc/vcs/DifftestEndpoint.sv
```diff
@@ -392,8 +397,7 @@ assign workload_switch = simv_result == `SIMV_GOODTRAP || simv_result == `SIMV_E
 assign difftest_uart_in_ch = 8'hff;
 always @(posedge clock) begin
   if (!reset && difftest_uart_out_valid) begin
-    $fwrite(32'h8000_0001, "%c", difftest_uart_out_ch);
-    $fflush();
+    difftest_uart_putc(difftest_uart_out_ch);
```
klin02: Maybe we should avoid using DPIC instead of system task(FWRITE), to reduce performance loss in Palladium.

klin02 (approved):

PR #870, opened 2026-05-06
chore(fpga): fallback release.sh without rsync

klin02:


klin02:
As `xs-env` container remove `rsync` from package list, this change add `tar` as  a fallback option for CI.

poemonsense (approved):

PR #872, opened 2026-05-07
chore: disable OBJCACHE for verilator by default

ngc7331:
1. it does not help XiangShan's build performance in most cases, as slight change in chisel result in significant change in cpp code
2. it introduces too much IO overhead

use 'make emu OBJCACHE=ccache' to enable it

See also: https://github.com/OpenXiangShan/XiangShan/pull/5907


ngc7331:
wait, this is not correct with remote build

poemonsense (approved):

PR #878, opened 2026-05-16
feat(difftest): generate checkpoint on difftest mismatch

good-circle:
Add an --auto-checkpoint option that asks the REF to dump a checkpoint when difftest reports a mismatch.

Without --enable-fork, generate the checkpoint after the mismatch is detected. With --enable-fork, wake the fork child and generate the checkpoint from the fork snapshot instead.

Route the request through RefProxy and pass a checkpoint base path to the REF.

### DIFF
--- a/src/test/csrc/common/args.cpp
+++ b/src/test/csrc/common/args.cpp
@@ -89,6 +89,7 @@ static inline void print_help(const char *file) {
   printf("      --as-footprints        load the image as memory access footprints\n");
   printf("      --dump-linearized=NAME dump the linearized footprints to NAME\n");
   printf("      --copy-ram=OFFSET      duplicate the memory at OFFSET\n");
+  printf("      --auto-checkpoint       save checkpoint to $NOOP_HOME/build/ on difftest error\n");
   printf("  -h, --help                 print program help info\n");
   printf("\n");
 }
@@ -98,6 +99,8 @@ CommonArgs parse_args(int argc, const char *argv[]) {
   int long_index = 0;
 #ifndef CONFIG_NO_DIFFTEST
   extern const char *difftest_ref_so;
+  extern bool difftest_auto_cpt;
+  extern bool cpt_from_fork;
 #endif // CONFIG_NO_DIFFTEST
 
   /* clang-format off */
@@ -133,6 +136,7 @@ CommonArgs parse_args(int argc, const char *argv[]) {
     { "instr-trace",       1, NULL,  0  },
     { "copy-ram",          1, NULL,  0  },
     { "cst-file",          1, NULL,  0  },
+    { "auto-checkpoint",   0, NULL,  0  },
     { "seed",              1, NULL, 's' },
     { "max-cycles",        1, NULL, 'C' },
     { "fork-interval",     1, NULL, 'X' },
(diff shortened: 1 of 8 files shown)


src/test/csrc/common/lightsss.cpp
```diff
@@ -23,6 +23,8 @@ ForkShareMemory::ForkShareMemory() {
   info->notgood = false;
   info->endCycles = 0;
   info->oldest = 0;
+  info->do_cpt = false;
```
poemonsense, line 26: There's no need to pass these. The child can decide it on its own.

src/test/csrc/common/args.cpp
```diff
@@ -98,6 +99,8 @@ CommonArgs parse_args(int argc, const char *argv[]) {
   int long_index = 0;
 #ifndef CONFIG_NO_DIFFTEST
   extern const char *difftest_ref_so;
+  extern bool difftest_auto_cpt;
+  extern bool cpt_from_fork;
```
poemonsense, line 103: Don't know why we need `cpt_from_fork`

src/test/csrc/common/args.cpp
```diff
@@ -167,7 +171,11 @@ CommonArgs parse_args(int argc, const char *argv[]) {
           case 4: difftest_ref_so = optarg; continue;
 #endif // CONFIG_NO_DIFFTEST
           case 5: args.enable_diff = false; continue;
-          case 6: args.enable_fork = true; continue;
+          case 6: args.enable_fork = true;
+#ifndef CONFIG_NO_DIFFTEST
+            cpt_from_fork = true;
```
poemonsense, line 176: Don't know why we need this.

src/test/csrc/difftest/refproxy.cpp
```diff
@@ -23,6 +23,8 @@
 
 uint8_t *ref_golden_mem = NULL;
 const char *difftest_ref_so = NULL;
+bool difftest_auto_cpt = false;
```
poemonsense, line 26: Don't know why we need this. Is it already in the args?

src/test/csrc/difftest/refproxy.h
```diff
@@ -468,6 +480,8 @@ struct InterruptDelegate {
 };
 
 extern const char *difftest_ref_so;
+extern bool difftest_auto_cpt;
```
poemonsense, line 483: Why do we need this

src/test/csrc/emu/emu.cpp
```diff
@@ -234,6 +234,12 @@ Emulator::~Emulator() {
   if (args.enable_fork && !is_fork_child()) {
     bool need_wakeup = trapCode != STATE_GOODTRAP && trapCode != STATE_LIMIT_EXCEEDED && trapCode != STATE_SIG;
     if (need_wakeup) {
+#ifndef CONFIG_NO_DIFFTEST
+      if (difftest_auto_cpt) {
```
poemonsense, line 238: No need to set the path here

src/test/csrc/emu/emu.cpp
```diff
@@ -758,5 +764,17 @@ void Emulator::fork_child_init() {
     difftest[i]->proxy->set_debug(true);
   }
 #endif
+
+  // Generate checkpoint from the fork snapshot state if requested
+  if (lightsss->get_do_cpt()) {
+    const char *base_filepath = lightsss->get_cpt_filepath();
```
poemonsense, line 770: The path should be determined here, the same as waveform.

src/test/csrc/emu/emu.cpp
```diff
@@ -758,5 +764,17 @@ void Emulator::fork_child_init() {
...
+    FORK_PRINTF("Generating checkpoint from fork snapshot: %s_memory_.gz\n", base_filepath)
+    for (int i = 0; i < NUM_CORES; i++) {
```
poemonsense, line 772: It seems we don't support multicore checkpoint here. Maybe we should check NUM_CORES and avoid checkpointing it.

poemonsense (commented):
It seems we don't need `difftest_auto_cpt` and `cpt_from_fork`

good-circle:
You are so right qaq

All these comments have been addressed. I will re-split the difftest changes and submit a new PR, and close this PR.

PR #879, opened 2026-05-21
feat(fpga): init random memory with AXI4

klin02:
This change randomizes hardware memory with seed/ramSize from xdma config, which is converted to same LFSR AXI4 as software, so FPGA and software memory can share same matching random memory. Note simMemory should clone full mem when random && seed != 0.

The memInit register in XDMA configBar has different values: 
  0 by default, 1 as WIP, 2 as Done, 3 as Fail (such as ramSize > addr).

Note DUT should expose CPU/Mem AXI4 IO to DifftestInterface, so SimTop can select between DUT CPU AXI4 and memory-init AXI4 from XDMA config. FPGA_SIM routes the AXI4 back to the exposed internal memory AXI4, while FPGA builds expose a standard difftest_mem AXI4 port for external DDR.

The XDMA config IO order:
1. set CfgReset 0 to reset configBar/memCtrl, disable CPU AXI.
2. random memInit with seed, use memInit AXI
3. set hostReset/memCPU to 1, use memCPU AXI


klin02:
Related PR:
https://github.com/OSCPU/NutShell/pull/258
https://github.com/OpenXiangShan/env-scripts/pull/128

xiaokamikami (dismissed):

xiaokamikami (approved):

PR #888, opened 2026-06-02
fix(SimTop): connext AXI in DUT when !isFPGA

klin02:


klin02:
DUT will connect AXI inside by default, and override the connection when isFPGA:
https://github.com/OSCPU/NutShell/pull/259

poemonsense (approved):

PR #891, opened 2026-06-04
feat(fpga): align 256bit AXIs and bank Diff2AXI for timing

klin02:
Use a 256-bit AXI-Stream interface to match the XDMA configuration (PCIe x4/x8 is 256-bit), and configure it through Gateway for both hardware parameters and software macros.
Also bank Diff2AXI with a 256-bit beat width to avoid wide memory and improve timing.


klin02:
FPGA SoC PR:
https://github.com/OpenXiangShan/env-scripts/pull/129

xiaokamikami (approved):

PR #898, opened 2026-06-21
feat(MMU): add ptehelper DPI model

fuhuakai:
Add a DPI-C page-table helper for XiangShan ptehelper simulation modes. The helper walks satp, vsatp, and hgatp page tables for noS2xlate, onlyStage1, onlyStage2, and allStage requests.

The model supports mPBMTE and hPBMTE, checks PBMT legality, keeps NAPOT PTE encoding visible to RTL, and reports VS-stage or G-stage page faults with stage-specific results. It also stops walks on faulting non-leaf PTEs and rejects out-of-range G-stage addresses.

Validated with XiangShan rvh-test in softTLB and softPTW modes. Both modes pass all rvh-test cases, including PBMT tests.

### DIFF
--- a/src/test/csrc/common/golden.cpp
+++ b/src/test/csrc/common/golden.cpp
@@ -23,31 +23,288 @@
 #include "perf.h"
 #endif // CONFIG_DIFFTEST_PERFCNT
 
-uint8_t pte_helper(uint64_t satp, uint64_t vpn, uint64_t *pte, uint8_t *level) {
+static constexpr uint8_t MODE_BARE = 0;
+static constexpr uint8_t MODE_SV39X4 = 8;
+static constexpr uint8_t MODE_SV48X4 = 9;
+static constexpr uint8_t VPN_BITS_PER_LEVEL = 9;
+static constexpr uint8_t GPADDR_BITS_SV39X4 = 41;
+static constexpr uint8_t GPADDR_BITS_SV48X4 = 50;
+
+/**
+ * Read PTE from memory (physical address)
+ * Uses goldenmem in difftest mode, pmem_read otherwise
+ */
+static inline uint64_t read_pte(uint64_t pte_addr) {
+  uint64_t pte_val;
+#ifdef CONFIG_NO_DIFFTEST
+  pte_val = pmem_read(pte_addr);
+#else
+  read_goldenmem(pte_addr, &pte_val, 8);
+#endif // CONFIG_NO_DIFFTEST
+  return pte_val;
+}
+
+static inline bool pte_is_leaf(const PTE &pte) {
+  return pte.v && (pte.r || pte.x || pte.w);
+}
+
+static inline bool pte_is_next(const PTE &pte) {
+  return pte.v && !pte.r && !pte.x && !pte.w;
+}
+
+static inline bool pte_has_pbmt_fault(const PTE &pte, bool pbmte) {
+  return pte.pbmt == 3 || (!pbmte && pte.pbmt != 0);
+}
+
+static inline bool pte_is_legal_napot(const PTE &pte, uint8_t level) {
+  return level == 0 && pte.n && (pte.ppn & 0xf) == 0x8;
+}
+
+static inline bool pte_has_napot_fault(const PTE &pte, uint8_t level) {
+  return pte.n && !pte_is_legal_napot(pte, level);
+}
+
+static inline bool pte_has_unaligned_fault(const PTE &pte, uint8_t level) {
+  if (!pte_is_leaf(pte) || level == 0) {
+    return false;
+  }
+  uint64_t ppn_mask = (1ull << (VPN_BITS_PER_LEVEL * level)) - 1;
+  return (pte.ppn & ppn_mask) != 0;
+}
+
+static inline bool pte_has_vs_stage_fault(const PTE &pte, uint8_t level, bool pbmte) {
+  if (pte.rsvd != 0 || pte_has_pbmt_fault(pte, pbmte)) {
+    return true;
+  }
+  if (pte_is_next(pte)) {
+    return pte.u || pte.a || pte.d || pte.n || pte.pbmt != 0;
+  }
+  return !pte.v || (!pte.r && pte.w) || (!pte_is_leaf(pte) && level == 0) || pte_has_napot_fault(pte, level) ||
+         pte_has_unaligned_fault(pte, level);
+}
+
+static inline bool pte_has_g_stage_fault(const PTE &pte, uint8_t level, bool pbmte) {
+  return pte_has_vs_stage_fault(pte, level, pbmte) || (pte_is_leaf(pte) && (!pte.u || !pte.a));
+}
+
+static inline bool gpa_high_bits_invalid(uint8_t mode, uint64_t gpaddr) {
+  if (mode == MODE_SV39X4) {
+    return (gpaddr >> GPADDR_BITS_SV39X4) != 0;
+  }
+  if (mode == MODE_SV48X4) {
+    return (gpaddr >> GPADDR_BITS_SV48X4) != 0;
+  }
+  return false;
+}
+
+static inline void set_pte_helper_result(uint64_t *pte, uint8_t *level, uint64_t *s1_pte, uint64_t *s2_pte,
+                                         uint8_t *s1_level, const PTE &result_pte, uint8_t result_level,
+                                         uint64_t result_s1_pte = 0, uint64_t result_s2_pte = 0,
+                                         uint8_t result_s1_level = 0) {
+  *pte = result_pte.val;
+  *level = result_level;
+  *s1_pte = result_s1_pte;
+  *s2_pte = result_s2_pte;
+  *s1_level = result_s1_level;
+}
+
+/**
+ * G-Stage page table walk (GPA -> HPA)
+ * Implementation aligned with do_s2xlate() in tlb.cpp
+ *
+ * @param hgatp  HGATP register value
+ * @param gpaddr Guest physical address to translate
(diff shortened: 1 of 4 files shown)


src/test/csrc/common/golden.cpp
```diff
@@ -23,31 +23,288 @@
...
+ * - onlyStage1: VS-stage only using vsatp (PTE addresses are physical)
+ * - onlyStage2: G-stage only using hgatp (vpn is GPA)
+ * - allStage:   Full two-stage (GVA -> GPA -> HPA)
+ */
+uint8_t pte_helper(uint64_t satp, uint64_t vsatp, uint64_t hgatp, uint8_t mPBMTE, uint8_t hPBMTE, uint64_t vpn,
+                   uint8_t s2xlate, uint64_t *pte, uint8_t *level, uint64_t *s1_pte, uint64_t *s2_pte,
+                   uint8_t *s1_level) {
+#ifdef CONFIG_DIFFTEST_PERFCNT
+  difftest_calls[perf_pte_helper]++;
+  difftest_bytes[perf_pte_helper] += 61;
```
fuhuakai, line 176: `difftest_bytes[perf_pte_helper]` follows the existing payload-byte
  counting rule. For the new `pte_helper` signature, the total is:
  `8 + 8 + 8 + 1 + 1 + 8 + 1 + 8 + 1 + 8 + 8 + 1 = 61`.

src/test/csrc/common/golden.cpp
```diff
@@ -23,31 +23,290 @@
...
+  }
+  if (pte_is_next(pte)) {
+    return pte.u || pte.a || pte.d || pte.n || pte.pbmt != 0;
+  }
+  return !pte.v || (!pte.r && pte.w) || (!pte_is_leaf(pte) && level == 0) || pte_has_napot_fault(pte, level) ||
+         pte_has_unaligned_fault(pte, level);
+}
+
+static inline bool pte_has_g_stage_fault(const PTE &pte, uint8_t level, bool pbmte) {
+  return pte_has_vs_stage_fault(pte, level, pbmte) || (pte_is_leaf(pte) && (!pte.u || !pte.a));
```
good-circle, line 89: Does pte.r need to be considered here? 

See https://github.com/OpenXiangShan/XiangShan/blob/5d393413277b3fab3043be145a1d48710586babf/src/main/scala/xiangshan/cache/mmu/PageTableWalker.scala#L409

src/test/csrc/common/golden.cpp
```diff
@@ -23,31 +23,290 @@
...
+          result_pte.val = 0;
+          set_pte_helper_result(pte, level, s1_pte, s2_pte, s1_level, result_pte, 0);
+          return PF_G_STAGE;
+        }
+
+        g_result = do_g_stage(hgatp_p, paddr, mPBMTE);
+
+        // Check for G-stage page fault during PTE access, including PBMTE-gated PBMT legality.
+        if (pte_has_g_stage_fault(g_result.pte, g_result.level, mPBMTE)) {
+          set_pte_helper_result(pte, level, s1_pte, s2_pte, s1_level, g_result.pte, g_result.level, 0, g_result.pte.val,
```
good-circle: For the `isForVSnonLeafPTE` case, i.e. when a G-stage translation is performed to support a VS-stage page-table access, if the G-stage translation faults, the value reported through `mtval2` / `htval` should correspond to the GPA of the implicit VS-stage page-table access (the GPA of the faulting VS-stage PTE).

In the current RTL, the L2 TLB/PTW response preserves the relevant S1 information and passes it back to the L1 TLB, where the final GPA is reconstructed using the S1 PTE/level information and the corresponding VPN index:
https://github.com/OpenXiangShan/XiangShan/blob/5d393413277b3fab3043be145a1d48710586babf/src/main/scala/xiangshan/cache/mmu/TLB.scala#L407

For this case, if `s1.pte` and `s1.level` are both returned as `0`, could this cause the L1 TLB to lose the information needed to reconstruct the faulting VS-stage PTE GPA, and therefore produce an incorrect `mtval2` / `htval`?

Also see https://github.com/OpenXiangShan/XiangShan/blob/5d393413277b3fab3043be145a1d48710586babf/src/main/scala/xiangshan/cache/mmu/PageTableWalker.scala#L252

src/test/csrc/common/golden.cpp
fuhuakai: Yes, I confirmed that this is a real permission-check bug.
I fixed this by adding a dedicated pte_has_g_stage_implicit_load_fault() check on the G-for-VS path. The generic G-stage fault check remains unchanged for onlyStage2 and the final GPA-to-HPA translation, whose R/W/X permissions are still handled according to the original request type. https://github.com/OpenXiangShan/difftest/pull/898/changes#diff-cf6b8e65e4f2be8ba884833b33404ad4c0718a5a7e42984222222e9d71b684bfR92-R94

src/test/csrc/common/golden.cpp
fuhuakai: Yes, a true bug. I think this is somehow similar to the idea of needgpa on the RTL.
I fixed this by maintaining s1_gpa_pte and s1_gpa_level throughout the VS-stage walk. Before the first VS PTE access, a fake PTE is initialized from vsatp.ppn to represent the VS root. After each successfully validated VS PTE, the saved PTE and level are updated. Any subsequent G-stage fault therefore returns the latest S1 context: root-level faults use the fake root PTE, while lower-level faults use the preceding non-leaf VS PTE and its level.
https://github.com/OpenXiangShan/difftest/pull/898/changes#diff-cf6b8e65e4f2be8ba884833b33404ad4c0718a5a7e42984222222e9d71b684bfR263-R265

klin02:
1. Maybe we can seperate these  `pte_helper`-related to another `pte.cpp`?
2. Have you test the verilog change before? Especially for `GSIM`, as the comb and reg logic may behave differently.

Also, I put XiangShan related PR here: https://github.com/OpenXiangShan/XiangShan/pull/6115

fuhuakai:
> 1. Maybe we can seperate these  `pte_helper`-related to another `pte.cpp`?
> 2. Have you test the verilog change before? Especially for `GSIM`, as the comb and reg logic may behave differently.
> 
> Also, I put XiangShan related PR here: [OpenXiangShan/XiangShan#6115](https://github.com/OpenXiangShan/XiangShan/pull/6115)

@klin02 Thank you for your suggestions above. Here are my own understandings and explanations.
1. Do you mean separating the two ideal DPI-C models currently implemented in golden.cpp, namely pte_helper and amo_helper, into independent modules (for example, pte.cpp/pte.h and amo.cpp/amo.h)? Or do you only suggest moving the page-table-related definitions and implementation into pte.cpp while keeping the AMO helper in golden.cpp?
My understanding is that the original file grouped them together because both are DPI-C-backed golden models, which made the current organization reasonably straightforward. If we split out only pte_helper,  I am not sure whether it would make theoverall organization clearer.
2. The current ptehelper implementation in this PR only supports the Verilator-based simulation flow. GSIM support has not been added yet. For the Verilator flow, I tested the ptehelper implementation with softTLB and softPTW enabled in their corresponding configurations, covering the RVH hypervisor tests, SPEC06 checkpoint workloads, and virtualized SPEC06 checkpoint workloads. These tests completed without observed functional bugs.

klin02:
> > 1. Maybe we can seperate these  `pte_helper`-related to another `pte.cpp`?
> > 2. Have you test the verilog change before? Especially for `GSIM`, as the comb and reg logic may behave differently.
> > 
> > Also, I put XiangShan related PR here: [OpenXiangShan/XiangShan#6115](https://github.com/OpenXiangShan/XiangShan/pull/6115)
> 
> @klin02 Thank you for your suggestions above. Here are my own understandings and explanations.
> 
> 1. Do you mean separating the two ideal DPI-C models currently implemented in golden.cpp, namely pte_helper and amo_helper, into independent modules (for example, pte.cpp/pte.h and amo.cpp/amo.h)? Or do you only suggest moving the page-table-related definitions and implementation into pte.cpp while keeping the AMO helper in golden.cpp?
>    My understanding is that the original file grouped them together because both are DPI-C-backed golden models, which made the current organization reasonably straightforward. If we split out only pte_helper,  I am not sure whether it would make theoverall organization clearer.
> 2. The current ptehelper implementation in this PR only supports the Verilator-based simulation flow. GSIM support has not been added yet. For the Verilator flow, I tested the ptehelper implementation with softTLB and softPTW enabled in their corresponding configurations, covering the RVH hypervisor tests, SPEC06 checkpoint workloads, and virtualized SPEC06 checkpoint workloads. These tests completed without observed functional bugs.

For 1, I think considering their common usage, it's OK to put these pte/amo helpers together as the PR is, but maybe we change the `golden` to a more clear name to clarify its usage. Does @good-circle have any suggestions?
For 2, I think the DPI model should be kept compatible for all our software backends, including VCS/Verilator/GSIM, as someone else may also use these DPI helpers. Maybe you can take a simple test with Gsim to see if this PR works as well? If there is any problem, we can also fix that with the GSIM team.

fuhuakai:
@klin02 Fine. I have already added GSIM support for this feature and ran virtualized SPEC CPU2006 checkpoints with GSIM for test. PteHelper now supports both Verilator and GSIM simulation backends. For both backends, softTLB and softPTW modes pass CI tests, including virtualized SPEC CPU2006 checkpoints. The pr has also been updated. https://github.com/OpenXiangShan/XiangShan/pull/6115/changes/e8002d2618bb38745d87e12745c19301dfe940d8#diff-c5bf21d167010715724f86039269a2a25af1b3da228e730b2b8e60f5919addd7R43-R75

good-circle (approved):

klin02 (approved):

PR #902, opened 2026-07-03
feat(flash): support compressed flash image

good-circle:
Flash checkpoints generated by NEMU can be stored as compressed flash images, such as _flash_.gz.

Load gzip/zstd flash images through the existing decompression helpers.

### DIFF
--- a/src/test/csrc/common/flash.cpp
+++ b/src/test/csrc/common/flash.cpp
@@ -16,6 +16,7 @@
 
 #include "flash.h"
 #include "common.h"
+#include "compress.h"
 #ifdef CONFIG_DIFFTEST_PERFCNT
 #include "perf.h"
 #endif // CONFIG_DIFFTEST_PERFCNT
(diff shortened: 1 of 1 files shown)


poemonsense (dismissed):

src/test/csrc/common/flash.cpp
```diff
@@ -91,6 +92,24 @@ void init_flash(const char *flash_bin) {
     eprintf(ANSI_COLOR_MAGENTA "[error] flash img not found\n");
     exit(1);
   }
+  fclose(flash_fp);
```
poemonsense: Should not use fclose here.

```
if (isGzFile ...) {
}

if (isZstdFile ...) {
}

// fallback here. Do not put this before isGzFile
FILE *flash_fp = fopen(flash_dev.img_path, "r");
  if (!flash_fp) {
    eprintf(ANSI_COLOR_MAGENTA "[error] flash img not found\n");
    exit(1);
  }
```

good-circle: fixed

poemonsense (approved):

PR #905, opened 2026-07-07
feat(topdown): collect IQ/ROB topdown with DPI-C

lewislzh:
- Add DPI-C helpers for topdown IQ and ROB info collection.
- Support both Verilator DPI and GSIM external-module paths.
- Share common IQ/ROB helper generation logic.


poemonsense:
@klin02  please check whether this PR could reuse the existing DifftestBundle definitions

klin02 (approved):

PR #917, opened 2026-07-17
feat(difftest): track vectorized matrix stores

Wonicon:
- Add vectorized  CStore event production and request/response source metadata
- Correlate all CStore req/resp by full encoded source ID per core
- Process CStore responses before requests and update golden memory exactly once
- Preserve pending CStores across replay and report correlation failures

### DIFF
--- a/src/main/scala/Bundles.scala
+++ b/src/main/scala/Bundles.scala
@@ -255,9 +255,13 @@ class SbufferEvent extends DifftestBaseBundle with HasValid {
 }
 
 class MatrixStoreEvent extends DifftestBaseBundle with HasValid {
+  val reqValid = Bool()
+  val reqEncodedSourceId = UInt(64.W)
   val addr = UInt(64.W)
   val data = Vec(64, UInt(8.W))
   val mask = UInt(64.W)
+  val respValid = Bool()
+  val respEncodedSourceId = UInt(64.W)
 }
 
 class UncacheMMStoreEvent extends DifftestBaseBundle with HasValid {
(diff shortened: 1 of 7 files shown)


yu-yake2002:
You can now submit pull requests directly to the master branch, which already supports the AME extension and has been adopted by XSAI. The feat-matrix branch is being retained for now because some colleagues are still using older versions of XSAI, which use the original feat-matrix branch.

Wonicon:
@copilot resolve the merge conflicts in this pull request

src/test/csrc/difftest/tests/matrix_store_tracker_test.cpp
yu-yake2002, line 1: The existing difftest doesn't seem to have this directory. If this test is unnecessary, I'd like to remove it.

Makefile
yu-yake2002, line 1: Makefiles should also be cleaned up.

yu-yake2002 (changes requested):

yu-yake2002:
@klin02 I'm not sure if the changes on `replay_snapshot` and `do_replay` are appropriate; could you please review them?

src/test/csrc/difftest/tests/matrix_store_tracker_test.cpp
Wonicon: > The existing difftest doesn't seem to have this directory. If this test is unnecessary, I'd like to remove it.

These are not necessary. I'll try a version without over-engineering

yu-yake2002 (approved):

PR #922, opened 2026-08-11
fix(difftest): preserve XiangShan version compatibility

poemonsense:
Split XiangShan into YQH, NH, KMHV2, and KMHV3 compatibility targets while keeping CPU_XIANGSHAN as the KMHV3 default.

Keep Zabha atomic handling exclusive to KMHV3 and select the updated flash initialization only for KMHV2 and KMHV3.

### DIFF
--- a/config/config.h
+++ b/config/config.h
@@ -20,7 +20,12 @@
 #include "difftest-state.h"
 
 #if defined(CPU_NUTSHELL)
+#elif defined(CPU_XIANGSHAN_YQH)
+#elif defined(CPU_XIANGSHAN_NH)
+#elif defined(CPU_XIANGSHAN_KMHV2)
+#elif defined(CPU_XIANGSHAN_KMHV3)
 #elif defined(CPU_XIANGSHAN)
+#define CPU_XIANGSHAN_KMHV3
 #elif defined(CPU_ROCKET_CHIP)
 #else
 // This is the default CPU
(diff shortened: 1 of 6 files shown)


poemonsense:
@Anzooooo @wissygh @good-circle @klin02 I'm adding some backward compatibility features to the repo so that prior XiangShan generations and XSAI can still be compatible with the latest DiffTest (cc @wakafa1). Please check whether the changes are good

Also, please help change the corresponding CPU String when bumping DiffTest next time @klin02

poemonsense:
The corresponding code changes are introduced in bd5a0273c375cf1c38b3f5ca10e347e095a7aef1 and 5d7d90bd6dcd7fede90183fc070f5e35897c6081

wissygh (approved):

wissygh:
https://github.com/OpenXiangShan/XiangShan/blob/f0624139063e12b87ad396144a5ddeaabc9c8a5c/src/main/scala/top/Top.scala#L454 Would it be better to change this to pass the value as a parameter?

wakafa1:
This version has passed testing on XSAI

poemonsense:
> https://github.com/OpenXiangShan/XiangShan/blob/f0624139063e12b87ad396144a5ddeaabc9c8a5c/src/main/scala/top/Top.scala#L454 Would it be better to change this to pass the value as a parameter?

It is a parameter generated by DiffTest. The XiangShan String will generate a C++ macro `CPU_XIANGSHAN`. @klin02 will update the string to XiangShan_KMHV3 when bumping difftest next time

PR #930, opened 2026-08-14
fix(runtime): support Darwin arm64 hosts

### LINKED ISSUE #929: [Feature Request] Support Darwin arm64 simulator hosts
**Related component**: build | simulation framework

**Is your feature request related to a problem? Please describe.**
The simulator runtime assumes Linux/glibc and x86 host APIs, so it cannot currently be built for Darwin arm64. Linux-only headers and dynamic-loader APIs are unavailable there, and some host type aliases differ.

**Describe the solution you'd like**
Make the host runtime platform-aware while preserving the existing Linux behavior: guard Linux-only APIs, use architecture-specific spin-wait instructions, use the Darwin dynamic-loader path, and use `size_t` consistently for memory byte counts.

**Describe alternatives you've considered**
Keeping an out-of-tree Apple-only patch works locally but duplicates maintenance and leaves the simulator runtime tied to one host platform.

**Additional context**
The intended first change is runtime-only. Cross-toolchain configuration, PGO, and performance tuning are outside its scope.

Gs-ygc:
The simulator runtime currently assumes Linux/glibc and x86 host APIs, which prevents it from being built for Darwin arm64.

Guard Linux-only headers, use architecture-specific spin-wait instructions, fall back to `dlopen()` on Darwin, and use `size_t` consistently for memory byte counts. Existing Linux behavior is unchanged.

Validated with the standalone Verilator build on Linux and native compilation with Apple Clang 17 on an Apple M4.

Closes #929.

### DIFF
--- a/src/test/csrc/common/lightsss.h
+++ b/src/test/csrc/common/lightsss.h
@@ -22,7 +22,9 @@
 #include <list>
 #include <signal.h>
 #include <sys/ipc.h>
+#if defined(__linux__)
 #include <sys/prctl.h>
+#endif
 #include <sys/shm.h>
 #include <sys/wait.h>
 #include <unistd.h>
--- a/src/test/csrc/common/mpool.h
+++ b/src/test/csrc/common/mpool.h
@@ -24,7 +24,9 @@
 #include <mutex>
 #include <stdexcept>
 #include <vector>
+#if defined(__x86_64__) || defined(__i386__)
 #include <xmmintrin.h>
+#endif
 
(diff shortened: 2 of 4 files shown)


poemonsense (approved):
I don't know much about arm64/darwin machines and cannot review whether the added code is good. But the changes look good to me.

eastonman:
I can confirm this works when compiling onto a RISC-V host.
I was trying to generate SimPoint checkpoint for XiangShan simulation binary using minjie-playground. And encountered a similar issue for the version used in minjie-playground. After bumping difftest version to include this PR, the checkpointing flow works fine.

PR #937, opened 2026-08-17
refactor(store.cpp): rewrite StoreRecorder::check for new store diff

huxuan0307:
This commit rewrite the function StoreRecorder::check
to full and 8B-aligned store events for difftest
in a simple way:

1. for case vecNeedSplit, seperate a store event into multiple eew-width parts.
    for case scalar, pass through as a whole.
2. for each commits, commits the lowhalf and highhalf seperately.

BREAKING CHANGE: store event generate is not backward compatible!

### DIFF
--- a/src/test/csrc/difftest/checkers/store.cpp
+++ b/src/test/csrc/difftest/checkers/store.cpp
@@ -15,11 +15,26 @@
 ***************************************************************************************/
 
 #include "checkers.h"
+#include "common.h"
 #include "diffstate.h"
 #include <cstdint>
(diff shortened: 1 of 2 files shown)


poemonsense (commented):
Similar to #922, please maintain backward compatibility by using macros and checker versions

Anzooooo (approved):

yu-yake2002:
This modification will cause compatibility issues in XSAI (`XIANGSHAN_KMHV2`). Please use macros to ensure compatibility.

PR #947, opened 2026-08-26
fix(build): honor BUILD_DIR for generated difftest artifacts

Gs-ygc:
## Summary

- derive `GENERATED_DIR` from `BUILD_DIR`
- write generated DiffTest artifacts below `GENERATED_DIR/generated-src`
- preserve the existing `<NOOP_HOME>/build` fallback

## Problem

The Verilator flow supports custom `BUILD_DIR` values for RTL and emulator outputs, but some Chisel and DiffTest artifacts are still written to `<NOOP_HOME>/build` and `<NOOP_HOME>/build/generated-src`.

This prevents parallel builds of different XiangShan configurations from the same `NOOP_HOME`. For example, `DefaultConfig` and `MinimalConfig`, or `DefaultMatrixConfig` with `ZhuJiang` and `OpenLLC`, can use different `BUILD_DIR` values while still overwriting shared generated files such as `DifftestMacros.svh`, `chisel_db.cpp`, `constantin.cpp`, and `perfCCT.cpp` under `<NOOP_HOME>/build`.

## Scope

This change only affects generated artifact placement. DiffTest runtime behavior is unchanged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,8 @@ export NOOP_HOME
 SIM_TOP    ?= SimTop
 DESIGN_DIR ?= $(NOOP_HOME)
 
-BUILD_DIR  = $(DESIGN_DIR)/build
+BUILD_DIR  ?= $(DESIGN_DIR)/build
+GENERATED_DIR ?= $(abspath $(BUILD_DIR))
 
 RTL_DIR = $(BUILD_DIR)/rtl
 RTL_SUFFIX ?= sv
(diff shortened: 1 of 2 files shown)


poemonsense (commented):
I don't think this is the correct way to implement this feature. A better alternative is to add a runtime argument for the build directory instead of override it with an environment variable.

I can merge this PR for now but I would re-implement this using a runtime argument in the near future. Please let me know whether we should merge this  (and refactor it in the near future) or improve it

Gs-ygc:
> I don't think this is the correct way to implement this feature. A better alternative is to add a runtime argument for the build directory instead of override it with an environment variable.
> 
> I can merge this PR for now but I would re-implement this using a runtime argument in the near future. Please let me know whether we should merge this (and refactor it in the near future) or improve it

You're right — using an environment variable to override the build directory isn't the cleanest approach. I'll switch it to a runtime argument instead. I'll revise the PR within the next two days. Thanks for the guidance!

klin02:
I think we should also seperate `GENERATED_DIR` for different `BUILD_DIR`, as in different config, difftest bundles/profile/... may have different instances.

PR #962, opened 2026-09-12
fix(ame): adapt packed events to NEMU ABI

yu-yake2002:
Batch mode emits packed Msync and AMU control events, while NEMU expects naturally aligned structs. Convert AME events in the checkers before calling the reference model while preserving the pointer-based RefProxy interface.


physics1024 (approved):
LGTM.

PR #965, opened 2026-09-14
chore(CI): switch xs-env containers to slim and drop GSIM install

klin02:
## Summary

Follow-up to [OpenXiangShan/xs-env#107](https://github.com/OpenXiangShan/xs-env/pull/107): gsim and clang-19 are now in the xs-env image, with `/usr/lib/llvm-19/bin` on `PATH`.

- replace `ghcr.io/openxiangshan/xs-env:ubuntu-*` with the `-slim` tags for smaller CI pulls
- drop the `test-gsim-rocketchip` GSIM download / `apt install clang-19` step
- drop `GSIM_CXX=clang++-19`; `gsim.mk` already defaults to `gsim` and `clang++`
- use `curl` instead of `wget` in nightly downloads (`wget` is optional and not in slim)

## Testing

- [ ] GitHub Actions CI on this PR


ngc7331 (approved):
LGTM, Thanks.

PR #967, opened 2026-09-15
feat(crob): Supports the new version of compression rob

HeiHuDie:


src/main/scala/Difftest.scala
```diff
@@ -631,6 +631,12 @@ object DifftestModule {
       difftest := DontCare
       difftest.bits.getValidOption.foreach(_ := false.B)
     }
+    difftest match {
+      // Existing producers have one exception boundary before their commits.
+      // A producer with paired slots explicitly overrides this for the latter.
+      case event: DiffArchEvent => event.isFormer := true.B
+      case _                    =>
+    }
```
klin02, line 639: Maybe you can use the DontCare(false.B) to set the ArchEvent fields. For example, use ArchEvent.nextSlot as !isFormer, and set nextSlot to false.B by default.

src/test/csrc/difftest/difftest.cpp
```diff
@@ -469,7 +469,9 @@ void Difftest::init_checkers() {
 #endif // CONFIG_DIFFTEST_LOADEVENT && CONFIG_DIFFTEST_SQUASH
 #ifdef CONFIG_DIFFTEST_STOREEVENT
   store_checker = new StoreChecker(state, proxy);
+#ifdef CONFIG_DIFFTEST_SQUASH
```
klin02, line 482: Why is this change related to store check? Is it necessary? If so, add some comment here.

src/test/csrc/difftest/difftest.cpp
```diff
@@ -649,30 +651,60 @@ inline int Difftest::check_all() {
...
       }
-    }
 #endif
-    for (int i = 0; i < CONFIG_DIFF_COMMIT_WIDTH; i++) {
-      if (dut->commit[i].valid) {
-        num_commit += 1 + dut->commit[i].nFused;
-        if (int ret = instr_commit_checker[i]->step()) {
-          return ret;
+    }
+    if (i == CONFIG_DIFF_COMMIT_WIDTH) {
```
klin02: Seems stange to use `i==XX` check in a `for(i=0 ; i<= XX; i++)` loop.
Maybe `for(i=0; i< XX; i++)` is enough?

src/main/scala/Difftest.scala
HeiHuDie: Agreed that nextSlot = !isFormer is a better encoding: false preserves the event-before-lane-0 behavior, while true places the event after the complete lane-0 commit, including its fused instructions. I propose renaming the field and updating both the producer and the consumer accordingly.

The current dontCare path invalidates the bundle rather than explicitly zero-initializing it, and the Chisel versions used here do not provide DontCare(false.B). We therefore need to guarantee an explicit false default. This also needs to cover legacy producers that overwrite the factory output with a whole-bundle connection, so I would check that path before removing the existing default handling.

src/test/csrc/difftest/difftest.cpp
HeiHuDie: This fixes a store-check scheduling issue exposed by CROB, independently of the new ArchEvent field. StoreRecorder enqueues the DUT store events before the commit loop. Without squash, StoreChecker drains that queue without an instruction stamp. If a batch contains a non-store former lane followed by a store latter lane, checking immediately after the former lane can reach the latter store before REF has executed it. The microbench case was addi sp, sp, -144 followed by sd ra, 136(sp).

The non-squash check therefore runs after the complete ordered batch. With squash, the check must remain per instruction because it uses commit_stamp to select store events. This also handles stores from valid commit lanes after an ArchEvent.

src/test/csrc/difftest/difftest.cpp
HeiHuDie: Agreed that this is harder to read than necessary. The extra iteration was intended to handle an event after lane 0 when the producer has only one commit lane. Simply changing <= to < while keeping the event check at the top of the loop would miss that boundary. However, this is a generic interface consideration; the current XiangShan configuration exposes 16 difftest commit lanes and does not need the extra iteration.

I propose using a normal < CONFIG_DIFF_COMMIT_WIDTH loop: process an interrupt or former event before the loop, and process a latter event immediately after completing lane 0. This preserves the order of fused instructions and any subsequent handler commits, handles the single-lane boundary naturally, and keeps the existing reference-PC synchronization before the first post-event commit.

poemonsense:
Please format the commit title/message to match the code changes.

The commit title/message should describe first what the commit changes, e.g., add nextSlot, then why we need this commit. Supporting CROB is why, not what.

poemonsense:
See https://chatgpt.com/share/6ab08ad5-c614-83ec-a351-5eb0cdf45f9f

PR #968, opened 2026-09-15
fix(verilator): disable SLP vectorization to avoid LLVM miscompilation

skyhgzsh:
> Avoid the X86 zero-mask/poison miscompilation tracked in LLVM #214388 on affected Clang releases. Keep other -O3 optimizations and quote OPT_FAST when passing multiple flags to recursive make.


Anzooooo (approved):

skyhgzsh:
Detail: when Clang19 compiling Cpp source from verilator, a Clang bug cause the emulator to compute wrong age matrix data, which triggered an assert.

PR #970, opened 2026-09-19
difftest: keep control trap markers running

good-circle:
NEMU control markers 0x100 and 0x101 are not simulation-exit codes. Map them to STATE_RUNNING so emu keeps going after nemu_trap.


klin02 (commented):
https://github.com/OpenXiangShan/difftest/blob/c41ad09261c6eca1f8424fa55e6dab31eaac482a/src/test/csrc/vcs/vcs_main.cpp#L366-L371
https://github.com/OpenXiangShan/difftest/blob/c41ad09261c6eca1f8424fa55e6dab31eaac482a/src/test/csrc/fpga/fpga_main.cpp#L270-L279
Maybe we should also change get_trapcode for vcs/fpga at the same time.

poemonsense (changes requested):
trap is a trap and should exit the simulation.

For co-simulation, the workload should not contain any traps until the end. If you are using workload-builder, should use `PROFILING=0` to create the Linux workloads if you are not taking checkpoints.

good-circle:
> trap is a trap and should exit the simulation.
> 
> For co-simulation, the workload should not contain any traps until the end. If you are using workload-builder, should use `PROFILING=0` to create the Linux workloads if you are not taking checkpoints.

Sure, I'll build workloads with PROFILING=0 and close this PR

